Issue #770, opened 2026-07-27
Provision NAS cluster access for the remote Kubernetes MCP


ppat:
## Summary

Manual, out-of-git provisioning needed to bring up `mcp-kubernetes-nas` (the read-only MCP for the `nas` cluster,
running on `homelab`) and to close out the `mcp-grafana` self-healing-token fix. Written in the style of
clusters#738 §3/§4 — this issue stays open after the associated PRs merge until the manual steps below are done
and verified.

Originates from apps#3370 / clusters#738 (initial rollout of the self-hosted MCP servers); companion to
ppat/homelab-ops-kubernetes-apps#3412 (`mcp-kubernetes-nas`) and ppat/homelab-ops-kubernetes-apps#3410 (the
`mcp-grafana` fix), and to #769 (the read-only ClusterRole added to both clusters).

## Progress

- [x] 1. Mint the nas ServiceAccount token
- [x] 2. Assemble and store the kubeconfig
- [x] 3. Grafana admin credential availability confirmed for the `ai` namespace
- [x] 4. Bump the `apps-ai` `ref.tag`
- [ ] 5. Post-rollout verification (mostly done — see §5; the Grafana pod-delete acceptance test and the
      NetworkPolicy debug-pod isolation check remain)
- [ ] 6. LiteLLM virtual-key updates + dead-key cleanup

## Landing sequence, and why the order matters

1. **Cluster-side RBAC first** (#769, landed via #771) — the `apps-ai` module (apps#3411) no longer ships its own
   ClusterRoleBinding; RBAC for `mcp-kubernetes-homelab` now lives exclusively in this repo. Bumping the `ref.tag`
   before this ClusterRole/ClusterRoleBinding exist on-cluster would leave `mcp-kubernetes-homelab` with zero
   permissions — no fallback binding ships with the module anymore.
2. **Manual credential/kubeconfig provisioning** (§1-§3 below) — the nas token, the assembled `kubeconfig_nas_mcp`,
   and confirmation that Grafana admin credentials are reachable from the `ai` namespace all need to exist before
   the new module version is live, since the Deployments reference them from first reconcile.
3. **Bump the `ref.tag`** (§4) — only safe once 1 and 2 are both in place.
4. **Post-rollout verification** (§5) — confirms the new RBAC, the nas MCP, and the Grafana self-healing token all
   work as designed, before anything depends on them.
[...]

PR #771, opened 2026-07-27
feat(kubernetes-api): read-only ClusterRole for the kubernetes MCP servers

### LINKED ISSUE #769: Add read-only ClusterRole for the Kubernetes MCP servers (homelab + nas)
## Summary

A purpose-built `mcp-kubernetes-readonly` ClusterRole is added to **both** clusters, in
`clusters/<cluster>/cluster/rbac/` alongside the existing `kubernetes-api-admin.yaml` and
`kubernetes-api-readonly.yaml` (which stay identical across `homelab` and `nas`), and bound to the
`mcp-kubernetes` ServiceAccounts that `apps-ai` ships namespaced-only. Unlike that existing pair, the nas variant
of this role deliberately diverges from homelab's — see below — since nas hosts the self-hosted Bitwarden instance
and carries a higher bar.

Originates from apps#3370 / clusters#738 (initial rollout of the self-hosted MCP servers); companion to
ppat/homelab-ops-kubernetes-apps#3411 (module-side `mcp-kubernetes` RBAC/toolset changes) and
ppat/homelab-ops-kubernetes-apps#3412 (the new `mcp-kubernetes-nas` server).

## Why this lives here, not in the module

This is cluster policy, not module behavior. The apps repo's own operating invariant is that cluster-wide coupling
lives at the edge — hard dependencies, cluster-scoped RBAC, and anything environment-specific are injected at the
point of consumption, never baked into a module. The `apps-ai` module therefore ships only the namespaced
`ServiceAccount` for each MCP server; the ClusterRole and ClusterRoleBinding that actually grant it anything belong
here, alongside the other cluster-wide RBAC that already lives in `clusters/<cluster>/cluster/rbac/`.

## What's granted, per cluster

- **homelab:** `mcp-kubernetes-readonly` ClusterRole + a ClusterRoleBinding to the existing `ai:mcp-kubernetes`
  ServiceAccount (already created by the `apps-ai` module).
- **nas:** a nas-specific variant of the ClusterRole (see "nas divergence" below), plus — since `nas` doesn't run
  `apps-ai` and has no in-cluster identity for this — a dedicated ServiceAccount in its own `mcp-access` namespace,
  its ClusterRoleBinding, and a `kubernetes.io/service-account-token` Secret to mint a long-lived token (Kubernetes
  1.24+ no longer auto-creates these for a ServiceAccount, so it has to be declared explicitly). The ServiceAccount
  is deliberately *not* placed in `kube-system`: that namespace is conventionally exempted from PSA enforcement and
  Kyverno namespace selectors, so an identity living there could silently inherit future exemptions. That token is
  what gets assembled into the kubeconfig consumed by `mcp-kubernetes-nas` — see #770.

## nas divergence

The nas role deliberately does not mirror homelab's byte-for-byte. nas hosts the self-hosted Bitwarden instance, so
it carries a higher bar. It drops `longhorn.io`, `tailscale.com`, `deviceplugin.intel.com` and `fpga.intel.com` —
none are deployed there (verified against `clusters/nas/kustomizations/`, which has no `*-extra` modules and no
`infra-storage-core`; nas storage is NFS + MinIO). It keeps `monitoring.coreos.com` (nas is getting prometheus
exporters plus promtail/alloy), `dragonflydb.io` and the CNPG groups (Harbor).

This sacrifices the byte-identical-across-clusters property that `kubernetes-api-admin`/`kubernetes-api-readonly`
have, and which aids review, in exchange for not silently granting read on an operator installed on nas later.
`configmaps` and `pods/log` remain granted cluster-wide on nas, including in the bitwarden namespace — a deliberate
accepted trade-off, since RBAC has no deny and excluding a single namespace would mean abandoning the ClusterRole
in favor of per-namespace Roles.

## Posture (restated here so this issue stands alone)

Read-only, no secret material or abuse potential, broad enough to be useful. RBAC is the primary control, not an
MCP-side config denylist (which cannot scale across the ~160 CRDs across 34 non-core API groups this cluster
carries — resource matching there is exact-string, no wildcards).

Granted API groups (homelab; nas differs per "nas divergence" above): `acme.cert-manager.io`, `cert-manager.io`,
`trust.cert-manager.io`, `external-secrets.io` (full), `generators.external-secrets.io` (explicit kind list, not
group-wide — see below), `externaldns.k8s.io`, the four `*.toolkit.fluxcd.io` groups, `postgresql.cnpg.io`,
`barmancloud.cnpg.io`, `longhorn.io`, `dragonflydb.io`, `monitoring.coreos.com`, `kyverno.io`, `policies.kyverno.io`,
`reports.kyverno.io`, `wgpolicyk8s.io`, `traefik.io`, `metallb.io` (minus `bgppeers`), `frrk8s.metallb.io`,
`gateway.networking.k8s.io`, `tailscale.com`, `autoscaling.k8s.io`, `deviceplugin.intel.com`, `fpga.intel.com`,
`nfd.k8s-sigs.io`, `upgrade.cattle.io`, `k3s.cattle.io`, plus all standard built-in groups and `metrics.k8s.io`.

Plus, explicitly:

- `nodes: list/get` — `nodes/proxy` is deliberately *not* granted. The API server maps HTTP method to RBAC verb for
  proxy subresources, and the kubelet registers GET routes — not only POST — for `/exec/`, `/attach/` and
  `/portForward/`, so `get` on `nodes/proxy` would authorise `GET /api/v1/nodes/<node>/proxy/exec/<ns>/<pod>/
  <container>?command=…` — exec-equivalent, not read-only, and the only rule that wouldn't be. Consequence:
  `nodes_log` and `nodes_stats_summary` do not work; `nodes_top` does — it needs only `nodes: list` plus
  `metrics.k8s.io` — which covers the node resource-usage need that motivated this work. The role's only
  non-`get`/`list`/`watch` rule is `selfsubjectaccessreviews: create`, next.
- `authorization.k8s.io/selfsubjectaccessreviews: create` — required for namespace defaulting in list operations.
  Deliberately *not* `subjectaccessreviews`/`localsubjectaccessreviews` or anything in `authentication.k8s.io` —
  those would let a caller probe other identities' RBAC and token validity, a materially different capability.
- `serviceaccounts: get/list` — names only; `ServiceAccount` objects haven't embedded tokens since Kubernetes 1.24.
- `rbac.authorization.k8s.io` read — aids debugging, confers no escalation on its own.
- `external-secrets.io` read (full) — specs reference secrets by name/`secretRef`, never by value.
- `generators.external-secrets.io` read, as an explicit kind allow-list rather than the whole group — the group-wide
  grant would contradict that same justification: `Fake.spec.data` is literal plaintext key/value pairs, and
  `Webhook.spec.headers` are free-form strings that can carry an inline API key, so both are excluded by name.
  `grafanas` is granted, with a stated trade-off: it exposes a plaintext Grafana admin username, the password
  staying in an unreadable Secret. The generated `Secret` objects any generator points at remain blocked.

Excluded, with reasons:

| Excluded | Why |
|---|---|
| core `secrets` | Credential material — the one thing this role must never expose |
| `helm.cattle.io` | `HelmChart.spec.valuesContent` is inline plaintext Helm values that can carry credentials; k3s-managed, outside repo audit boundary |
| `hub.traefik.io` | Unlicensed/unused; `APIAuth`/`AccessControlPolicy` hold auth config — zero value, nonzero risk |
| `monitoring.grafana.com` | Grafana Agent Operator — unused |
| `metallb.io/bgppeers` specifically | `BGPPeer.spec.password` is a plaintext BGP password; both clusters are L2-only, so nothing operational is lost by excluding just this one resource kind |
| `authentication.k8s.io`, non-self access-review kinds | Token-validation and cross-identity RBAC oracles |
| `generators.external-secrets.io/fakes`, `.../webhooks` specifically | `Fake.spec.data` is literal plaintext key/value pairs; `Webhook.spec.headers` are free-form strings that can carry an inline API key |
| `nodes/proxy` | Verb-to-route mapping makes `get` exec-equivalent for the kubelet's `/exec/`, `/attach/`, `/portForward/` routes — see above |

Documented residual risks, accepted and revisitable: `configmaps`/pod specs can disclose Secret *names* and mount
paths (an audit found no inlined credentials today, but that's a discipline guarantee, not a technical one);
`traefik.io/Middleware`'s inline surface is `spec.headers.customRequestHeaders` (can carry an
`Authorization: Bearer …` value) and the free-form `spec.plugin` config, not `spec.basicAuth` (that field only ever
takes a `secret` reference on this CRD, never inline credentials); `monitoring.coreos.com/AlertmanagerConfig` can
embed webhook URLs; on homelab, `tailscale.com/ProxyClass.spec.statefulSet.pod.tailscaleContainer.env` accepts
inline env vars, `TS_AUTHKEY` being the obvious one (not applicable to nas, which drops the group — see "nas
divergence" above).

## Also in scope

Bump the `apps-ai` `GitRepository` `ref.tag` (in `clusters/<cluster>/sources/`) once the apps-repo release
containing the `mcp-kubernetes` RBAC/toolset changes (ppat/homelab-ops-kubernetes-apps#3411) and the new
`mcp-kubernetes-nas` module (ppat/homelab-ops-kubernetes-apps#3412) is cut. That release is now published
(`apps-ai-v0.6.10`); the tag bump itself is tracked as part of the outstanding rollout sequence in #770, since
bumping it is only safe once this ClusterRole/binding exist on both clusters (they now do).

## Out of scope / deferred

- Minting the `nas` long-lived token, assembling the kubeconfig, and the rest of the manual provisioning for
  `mcp-kubernetes-nas` — tracked in #770.

ppat:
Closes #769. Cluster side of ppat/homelab-ops-kubernetes-apps#3411 and ppat/homelab-ops-kubernetes-apps#3412.

## Why

`mcp-kubernetes` was bound to the stock `view` ClusterRole. `view` excludes all cluster-scoped resources and only picks up CRDs whose operators set the `rbac.authorization.k8s.io/aggregate-to-view` label — so on a cluster running ~160 CRDs it could not read nodes, node metrics, PersistentVolumes, StorageClasses, CRDs, APIServices, or anything from CloudNativePG, Longhorn, Prometheus Operator, MetalLB, Traefik, Tailscale or VPA.

Root of the design: use standard Kubernetes RBAC as the primary control instead of the MCP server's own `denied_resources` config. That config matches Group/Version/Kind by exact string equality with no wildcard support, so it cannot scale across this many CRDs.

## What

A purpose-built `mcp-kubernetes-readonly` ClusterRole — an explicit read-only allow-list over the standard built-in groups, `metrics.k8s.io`, and 29 CRD groups.

Verbs never exceed `get/list/watch`, with exactly one exception:

| Exception | Verb | Why |
|---|---|---|
| `selfsubjectaccessreviews` | `create` | Required by the server's `resources_list` for namespace defaulting. Deliberately excludes `subjectaccessreviews`/`localsubjectaccessreviews` and all of `authentication.k8s.io`, which are RBAC and token-validation oracles. |

`nodes/proxy` is deliberately **not** granted, despite motivating this work. The API server maps HTTP method to RBAC verb for proxy subresources, and the kubelet registers GET routes — not only POST — for `/exec/`, `/attach/` and `/portForward/`. So `get` on `nodes/proxy` would authorise `GET /api/v1/nodes/<node>/proxy/exec/<ns>/<pod>/<container>?command=…`, i.e. it's exec-equivalent, not read-only — it would have been the only non-read-only rule in the role. Consequence: `nodes_log` and `nodes_stats_summary` don't work; `nodes_top` does — it only needs `nodes: list` plus `metrics.k8s.io` — which covers the node resource-usage visibility that motivated this work.

Core `secrets` are absent from every rule. Also excluded, each documented inline where it matters:

- `helm.cattle.io` — `HelmChart.spec.valuesContent` is inline plaintext Helm values that can carry credentials
- `hub.traefik.io` — Traefik Hub, unlicensed/unused; `APIAuth`/`AccessControlPolicy` hold auth config
- `monitoring.grafana.com` — Grafana Agent Operator, unused
- `metallb.io/bgppeers` — `BGPPeer.spec.password` is a plaintext BGP password; MetalLB runs L2-only here, so the group is granted by explicit resource list instead of `*`
- `generators.external-secrets.io/fakes`, `.../webhooks` — `Fake.spec.data` is literal plaintext key/value pairs; `Webhook.spec.headers` are free-form strings that can carry an inline API key

`external-secrets.io` is fully readable: its specs reference secrets by name and never contain values, and the generated Secrets themselves remain blocked. `generators.external-secrets.io` is readable only as an explicit kind list rather than group-wide, per the exclusions above; it does include `grafanas`, with a stated trade-off — it exposes a plaintext Grafana admin username, the password staying in an unreadable Secret.

## Placement

This lives here rather than in the module because a cluster-wide grant is cluster policy, matching the existing `kubernetes-api-admin.yaml` / `kubernetes-api-readonly.yaml` convention. The apps-ai module now ships only its namespaced ServiceAccount.

## nas cluster

nas gets its own variant of the ClusterRole rather than a copy of homelab's, and deliberately so — nas hosts the self-hosted Bitwarden instance, so it carries a higher bar. It drops `longhorn.io`, `tailscale.com`, `deviceplugin.intel.com` and `fpga.intel.com` (nothing on nas uses them — nas is NFS + MinIO, no `*-extra` modules, no `infra-storage-core`), while keeping `monitoring.coreos.com` (prometheus exporters plus promtail/alloy are landing there), `dragonflydb.io` and the CNPG groups (Harbor). This trades away the byte-identical-across-clusters property the existing `kubernetes-api-admin`/`kubernetes-api-readonly` convention has — which aids review — for not silently granting read on an operator installed on nas later. `configmaps` and `pods/log` stay granted cluster-wide on nas, including in the bitwarden namespace: a deliberate accepted trade-off, since RBAC has no deny and excluding one namespace would mean abandoning the ClusterRole for per-namespace Roles.

nas additionally gets a dedicated ServiceAccount — in its own `mcp-access` namespace, not `kube-system`, since `kube-system` is conventionally exempted from PSA enforcement and Kyverno namespace selectors and could silently inherit future exemptions — and a long-lived `kubernetes.io/service-account-token` Secret, so the remote MCP server running on homelab can authenticate against the nas API server.

## Accepted residual risks

Documented so they stay revisitable: `configmaps` and pod specs disclose Secret names and mount paths, and would expose any credential a future contributor inlines — an audit of both repos found none today, but that is a discipline guarantee, not a technical one. `traefik.io/Middleware`'s inline surface is `spec.headers.customRequestHeaders` (can carry an `Authorization: Bearer …` value) and the free-form `spec.plugin` config — not `spec.basicAuth`, which on this CRD only ever takes a `secret` reference, never inline credentials. `monitoring.coreos.com/AlertmanagerConfig` can embed webhook URLs. On homelab, `tailscale.com/ProxyClass.spec.statefulSet.pod.tailscaleContainer.env` accepts inline env vars, `TS_AUTHKEY` being the obvious one (not applicable to nas, which drops the group).

## Still manual (tracked in #770)

Minting the nas token, assembling the kubeconfig into Bitwarden as `kubeconfig_nas_mcp`, and confirming the nas API server address resolves from inside a homelab pod. Automated rotation of the nas token is deliberately deferred until there's a consistent rotation mechanism worth building; revocation is `kubectl delete secret` and effective immediately. With `nodes/proxy` dropped, this token is genuinely read-only.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -106,6 +106,58 @@ jobs:
     with:
       git_ref: ${{ github.head_ref || github.ref }}
 
+  rbac-clusterrole-readonly:
+    needs: [detect-changes]
+    if: ${{ github.event_name != 'pull_request' || fromJSON(needs.detect-changes.outputs.results).kubernetes_any_changed == 'true' }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    env:
+      # Pinned to the Kyverno CLI release matching the appVersion of the
+      # kyverno chart deployed by infra-security-core (see the apps repo's
+      # infrastructure/subsystems/security-core/kyverno/helm-release-kyverno.yaml),
+      # so the CLI's policy engine matches what's actually running.
+      KYVERNO_CLI_VERSION: v1.18.2
+      KYVERNO_CLI_SHA256: cb2feb8356149fd2fe774c894ccf0969f4a60a83867dd913af724f74ffbbc18b
+    steps:
+    - name: Checkout
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        fetch-depth: 1
+        persist-credentials: false
+
+    - name: Install Kyverno CLI
+      shell: bash
+      # yamllint disable-line rule:indentation
+      run: |
+        set -euo pipefail
+        curl -sL -o /tmp/kyverno-cli.tar.gz \
+          "https://github.com/kyverno/kyverno/releases/download/${KYVERNO_CLI_VERSION}/kyverno-cli_${KYVERNO_CLI_VERSION}_linux_x86_64.tar.gz"
+        echo "${KYVERNO_CLI_SHA256}  /tmp/kyverno-cli.tar.gz" | sha256sum -c -
+        mkdir -p "${RUNNER_TEMP}/kyverno-cli"
+        tar -xzf /tmp/kyverno-cli.tar.gz -C "${RUNNER_TEMP}/kyverno-cli" kyverno
+        echo "${RUNNER_TEMP}/kyverno-cli" >> "$GITHUB_PATH"
+
+    - name: Validate RBAC ClusterRoles are read-only (Kyverno CLI, no cluster required)
+      shell: bash
+      # yamllint disable-line rule:indentation
+      run: |
+        set -uo pipefail
+        # One `kyverno apply` invocation per cluster directory: every cluster
+        # defines a ClusterRole named mcp-kubernetes-readonly, and since
+        # ClusterRoles are cluster-scoped (no namespace to disambiguate them),
+        # loading two clusters' rbac/ directories into a single invocation
+        # makes the CLI's resource cache collide same-named resources from
+        # different clusters together, silently dropping one.
+        status=0
+        for dir in clusters/*/cluster/rbac; do
+          echo "::group::${dir}"
+          if ! kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml --resource "${dir}/" --detailed-results; then
+            status=1
+          fi
+          echo "::endgroup::"
+        done
+        exit "${status}"
+
   renovate-config-check:
     needs: [detect-changes]
     if: ${{ github.event_name != 'pull_request' || fromJSON(needs.detect-changes.outputs.results).renovate_any_changed == 'true' }}
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -106,58 +106,6 @@ jobs:
     with:
       git_ref: ${{ github.head_ref || github.ref }}
 
-  rbac-clusterrole-readonly:
-    needs: [detect-changes]
-    if: ${{ github.event_name != 'pull_request' || fromJSON(needs.detect-changes.outputs.results).kubernetes_any_changed == 'true' }}
-    runs-on: ubuntu-24.04
-    timeout-minutes: 5
-    env:
-      # Pinned to the Kyverno CLI release matching the appVersion of the
-      # kyverno chart deployed by infra-security-core (see the apps repo's
-      # infrastructure/subsystems/security-core/kyverno/helm-release-kyverno.yaml),
-      # so the CLI's policy engine matches what's actually running.
-      KYVERNO_CLI_VERSION: v1.18.2
-      KYVERNO_CLI_SHA256: cb2feb8356149fd2fe774c894ccf0969f4a60a83867dd913af724f74ffbbc18b
-    steps:
-    - name: Checkout
-      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
-      with:
-        fetch-depth: 1
-        persist-credentials: false
-
-    - name: Install Kyverno CLI
-      shell: bash
-      # yamllint disable-line rule:indentation
-      run: |
-        set -euo pipefail
-        curl -sL -o /tmp/kyverno-cli.tar.gz \
-          "https://github.com/kyverno/kyverno/releases/download/${KYVERNO_CLI_VERSION}/kyverno-cli_${KYVERNO_CLI_VERSION}_linux_x86_64.tar.gz"
-        echo "${KYVERNO_CLI_SHA256}  /tmp/kyverno-cli.tar.gz" | sha256sum -c -
-        mkdir -p "${RUNNER_TEMP}/kyverno-cli"
-        tar -xzf /tmp/kyverno-cli.tar.gz -C "${RUNNER_TEMP}/kyverno-cli" kyverno
-        echo "${RUNNER_TEMP}/kyverno-cli" >> "$GITHUB_PATH"
-
-    - name: Validate RBAC ClusterRoles are read-only (Kyverno CLI, no cluster required)
-      shell: bash
-      # yamllint disable-line rule:indentation
-      run: |
-        set -uo pipefail
-        # One `kyverno apply` invocation per cluster directory: every cluster
-        # defines a ClusterRole named mcp-kubernetes-readonly, and since
-        # ClusterRoles are cluster-scoped (no namespace to disambiguate them),
-        # loading two clusters' rbac/ directories into a single invocation
-        # makes the CLI's resource cache collide same-named resources from
-        # different clusters together, silently dropping one.
-        status=0
-        for dir in clusters/*/cluster/rbac; do
-          echo "::group::${dir}"
-          if ! kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml --resource "${dir}/" --detailed-results; then
-            status=1
-          fi
-          echo "::endgroup::"
-        done
-        exit "${status}"
-
   renovate-config-check:
     needs: [detect-changes]
     if: ${{ github.event_name != 'pull_request' || fromJSON(needs.detect-changes.outputs.results).renovate_any_changed == 'true' }}
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -2,11 +2,13 @@
 # yamllint disable rule:line-length
 # Static analysis of security invariants — distinct from lint.yaml (which
 # only checks style/formatting via yamllint/markdownlint/commitlint/etc).
-# rbac-clusterroles statically validates the read-only RBAC ClusterRoles in
-# clusters/*/cluster/rbac/ against ci/policy-tests/rbac-clusterrole-readonly.yaml
-# using the Kyverno CLI (`kyverno apply`, no cluster required). That policy
-# is test-only: it lives under ci/ and, unlike policies/**, is never wired
-# up to a Flux Kustomization or applied to a cluster.
+# rbac-clusterroles statically validates the read-only RBAC ClusterRoles and
+# their ClusterRoleBindings in clusters/*/cluster/rbac/ against the two
+# policies in ci/policy-tests/ (rbac-clusterrole-readonly.yaml and
+# rbac-clusterrolebinding-roleref.yaml) using the Kyverno CLI (`kyverno
+# apply`, no cluster required). Those policies are test-only: they live
+# under ci/ and, unlike policies/**, are never wired up to a Flux
+# Kustomization or applied to a cluster.
 name: static-analysis
 
 on:
@@ -75,7 +77,7 @@ jobs:
           # just a pass/fail summary count. --remove-color strips ANSI codes
           # that would otherwise print as raw escape sequences in the
           # non-interactive Actions log.
-          if ! mise exec -- kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml --resource "${dir}/" --table --detailed-results --remove-color; then
+          if ! mise exec -- kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml ci/policy-tests/rbac-clusterrolebinding-roleref.yaml --resource "${dir}/" --table --detailed-results --remove-color; then
             status=1
           fi
           echo "::endgroup::"
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -66,7 +66,16 @@ jobs:
         status=0
         for dir in clusters/*/cluster/rbac; do
           echo "::group::${dir}"
-          if ! mise exec -- kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml --resource "${dir}/" --detailed-results; then
+          echo "Resource files loaded from ${dir}:"
+          find "${dir}" -maxdepth 1 -type f -name '*.yaml' | sort
+          echo
+          # --table + --detailed-results renders one row per (rule, resource)
+          # pair with its pass/fail outcome and message, so the log itself
+          # shows every rule that ran against every matched resource, not
+          # just a pass/fail summary count. --remove-color strips ANSI codes
+          # that would otherwise print as raw escape sequences in the
+          # non-interactive Actions log.
+          if ! mise exec -- kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml --resource "${dir}/" --table --detailed-results --remove-color; then
             status=1
           fi
           echo "::endgroup::"
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -38,28 +38,21 @@ jobs:
       # kyverno chart deployed by infra-security-core (see the apps repo's
       # infrastructure/subsystems/security-core/kyverno/helm-release-kyverno.yaml),
       # so the CLI's policy engine matches what's actually running.
-      KYVERNO_CLI_VERSION: v1.18.2
-      KYVERNO_CLI_SHA256: cb2feb8356149fd2fe774c894ccf0969f4a60a83867dd913af724f74ffbbc18b
+      # renovate: datasource=github-releases depName=kyverno/kyverno
+      KYVERNO_CLI_VERSION: "v1.18.2"
     steps:
-    - name: Checkout
-      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+    - name: Setup repository and tools
+      uses: ppat/homelab-ops-actions/actions/setup-repository-tools@f1ae73db4ab080b7e8dc66d078e5905d5ca526cb # v2.1.0
       with:
-        fetch-depth: 1
-        persist-credentials: false
-
-    - name: Install Kyverno CLI
-      shell: bash
-      # yamllint disable-line rule:indentation
-      run: |
-        set -euo pipefail
-        curl -sL -o /tmp/kyverno-cli.tar.gz \
-          "https://github.com/kyverno/kyverno/releases/download/${KYVERNO_CLI_VERSION}/kyverno-cli_${KYVERNO_CLI_VERSION}_linux_x86_64.tar.gz"
-        echo "${KYVERNO_CLI_SHA256}  /tmp/kyverno-cli.tar.gz" | sha256sum -c -
-        mkdir -p "${RUNNER_TEMP}/kyverno-cli"
-        tar -xzf /tmp/kyverno-cli.tar.gz -C "${RUNNER_TEMP}/kyverno-cli" kyverno
-        echo "${RUNNER_TEMP}/kyverno-cli" >> "$GITHUB_PATH"
+        current_git_ref: ${{ github.head_ref || github.ref }}
+        current_repository: ${{ github.repository }}
+        token: ${{ secrets.GITHUB_TOKEN }}
+        mise_toml: |
+          [tools]
+          kyverno = "${{ env.KYVERNO_CLI_VERSION }}"
 
     - name: Validate RBAC ClusterRoles are read-only (Kyverno CLI, no cluster required)
+      working-directory: ./current
       shell: bash
       # yamllint disable-line rule:indentation
       run: |
@@ -73,7 +66,7 @@ jobs:
         status=0
         for dir in clusters/*/cluster/rbac; do
           echo "::group::${dir}"
-          if ! kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml --resource "${dir}/" --detailed-results; then
+          if ! mise exec -- kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml --resource "${dir}/" --detailed-results; then
             status=1
           fi
           echo "::endgroup::"
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -13,9 +13,11 @@ on:
   pull_request:
     paths:
     - '.github/workflows/static-analysis.yaml'
-    - 'clusters/**/*.yaml'
-    - 'ci/**/*.yaml'
-    - 'policies/**/*.yaml'
+    # Scoped to exactly what this analyses and what does the analysing -- the RBAC
+    # manifests under test and the policy asserting their invariants. Broader globs would
+    # run it on every clusters/ or ci/ edit for no signal.
+    - 'clusters/*/cluster/rbac/**'
+    - 'ci/policy-tests/**'
   schedule:
   - cron: '0 5 * * 1'
   workflow_dispatch:
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -1,0 +1,79 @@
+---
+# yamllint disable rule:line-length
+# Static analysis of security invariants — distinct from lint.yaml (which
+# only checks style/formatting via yamllint/markdownlint/commitlint/etc).
+# rbac-clusterroles statically validates the read-only RBAC ClusterRoles in
+# clusters/*/cluster/rbac/ against ci/policy-tests/rbac-clusterrole-readonly.yaml
+# using the Kyverno CLI (`kyverno apply`, no cluster required). That policy
+# is test-only: it lives under ci/ and, unlike policies/**, is never wired
+# up to a Flux Kustomization or applied to a cluster.
+name: static-analysis
+
+on:
+  pull_request:
+    paths:
+    - '.github/workflows/static-analysis.yaml'
+    - 'clusters/**/*.yaml'
+    - 'ci/**/*.yaml'
+    - 'policies/**/*.yaml'
+  schedule:
+  - cron: '0 5 * * 1'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  rbac-clusterroles:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    env:
+      # Pinned to the Kyverno CLI release matching the appVersion of the
+      # kyverno chart deployed by infra-security-core (see the apps repo's
+      # infrastructure/subsystems/security-core/kyverno/helm-release-kyverno.yaml),
+      # so the CLI's policy engine matches what's actually running.
+      KYVERNO_CLI_VERSION: v1.18.2
+      KYVERNO_CLI_SHA256: cb2feb8356149fd2fe774c894ccf0969f4a60a83867dd913af724f74ffbbc18b
+    steps:
+    - name: Checkout
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        fetch-depth: 1
+        persist-credentials: false
+
+    - name: Install Kyverno CLI
+      shell: bash
+      # yamllint disable-line rule:indentation
+      run: |
+        set -euo pipefail
+        curl -sL -o /tmp/kyverno-cli.tar.gz \
+          "https://github.com/kyverno/kyverno/releases/download/${KYVERNO_CLI_VERSION}/kyverno-cli_${KYVERNO_CLI_VERSION}_linux_x86_64.tar.gz"
+        echo "${KYVERNO_CLI_SHA256}  /tmp/kyverno-cli.tar.gz" | sha256sum -c -
+        mkdir -p "${RUNNER_TEMP}/kyverno-cli"
+        tar -xzf /tmp/kyverno-cli.tar.gz -C "${RUNNER_TEMP}/kyverno-cli" kyverno
+        echo "${RUNNER_TEMP}/kyverno-cli" >> "$GITHUB_PATH"
+
+    - name: Validate RBAC ClusterRoles are read-only (Kyverno CLI, no cluster required)
+      shell: bash
+      # yamllint disable-line rule:indentation
+      run: |
+        set -uo pipefail
+        # One `kyverno apply` invocation per cluster directory: every cluster
+        # defines a ClusterRole named mcp-kubernetes-readonly, and since
+        # ClusterRoles are cluster-scoped (no namespace to disambiguate them),
+        # loading two clusters' rbac/ directories into a single invocation
+        # makes the CLI's resource cache collide same-named resources from
+        # different clusters together, silently dropping one.
+        status=0
+        for dir in clusters/*/cluster/rbac; do
+          echo "::group::${dir}"
+          if ! kyverno apply ci/policy-tests/rbac-clusterrole-readonly.yaml --resource "${dir}/" --detailed-results; then
+            status=1
+          fi
+          echo "::endgroup::"
+        done
+        exit "${status}"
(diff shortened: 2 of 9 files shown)


ppat:
`````diff
--- clusters/homelab Kustomization: flux-system/root ClusterRole: flux-system/mcp-kubernetes-readonly

+++ clusters/homelab Kustomization: flux-system/root ClusterRole: flux-system/mcp-kubernetes-readonly

@@ -0,0 +1,302 @@

+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  annotations:
+    kustomize.toolkit.fluxcd.io/prune: disabled
+  labels:
+    kustomize.toolkit.fluxcd.io/name: root
+    kustomize.toolkit.fluxcd.io/namespace: flux-system
+  name: mcp-kubernetes-readonly
+rules:
+- apiGroups:
+  - ''
+  resources:
+  - configmaps
+  - endpoints
+  - events
+  - limitranges
+  - namespaces
+  - nodes
+  - nodes/status
+  - persistentvolumeclaims
+  - persistentvolumes
+  - pods
+  - pods/log
+  - pods/status
+  - podtemplates
+  - replicationcontrollers
+  - resourcequotas
+  - serviceaccounts
+  - services
+  - services/status
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - controllerrevisions
+  - daemonsets
+  - daemonsets/status
+  - deployments
+  - deployments/scale
+  - deployments/status
+  - replicasets
+  - replicasets/scale
+  - replicasets/status
+  - statefulsets
+  - statefulsets/scale
+  - statefulsets/status
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - batch
+  resources:
+  - cronjobs
+  - cronjobs/status
+  - jobs
+  - jobs/status
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - autoscaling
+  resources:
+  - horizontalpodautoscalers
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - policy
+  resources:
+  - poddisruptionbudgets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - events.k8s.io
+  resources:
+  - events
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - ingresses
+  - ingressclasses
+  - networkpolicies
+  - ipaddresses
+  - servicecidrs
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - csidrivers
+  - csinodes
+  - csistoragecapacities
+  - storageclasses
+  - volumeattachments
+  - volumeattributesclasses
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - scheduling.k8s.io
+  resources:
+  - priorityclasses
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - node.k8s.io
+  resources:
+  - runtimeclasses
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - apiregistration.k8s.io
+  resources:
+  - apiservices
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - mutatingwebhookconfigurations
+  - validatingwebhookconfigurations
+  - mutatingadmissionpolicies
+  - mutatingadmissionpolicybindings
+  - validatingadmissionpolicies
+  - validatingadmissionpolicybindings
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - certificates.k8s.io
+  resources:
+  - certificatesigningrequests
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - flowcontrol.apiserver.k8s.io
+  resources:
+  - flowschemas
+  - prioritylevelconfigurations
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - resource.k8s.io
+  resources:
+  - deviceclasses
+  - resourceclaims
+  - resourceclaimtemplates
+  - resourceslices
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
+  - clusterroles
+  - clusterrolebindings
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - metrics.k8s.io
+  resources:
+  - nodes
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - authorization.k8s.io
+  resources:
+  - selfsubjectaccessreviews
+  verbs:
+  - create
+- apiGroups:
+  - acme.cert-manager.io
+  - cert-manager.io
+  - trust.cert-manager.io
+  - external-secrets.io
+  - externaldns.k8s.io
+  - source.toolkit.fluxcd.io
+  - kustomize.toolkit.fluxcd.io
+  - helm.toolkit.fluxcd.io
+  - notification.toolkit.fluxcd.io
+  - postgresql.cnpg.io
+  - barmancloud.cnpg.io
+  - longhorn.io
+  - dragonflydb.io
+  - monitoring.coreos.com
+  - kyverno.io
+  - policies.kyverno.io
+  - reports.kyverno.io
+  - wgpolicyk8s.io
+  - traefik.io
+  - frrk8s.metallb.io
+  - gateway.networking.k8s.io
+  - tailscale.com
+  - autoscaling.k8s.io
+  - deviceplugin.intel.com
+  - fpga.intel.com
+  - nfd.k8s-sigs.io
+  - upgrade.cattle.io
+  - k3s.cattle.io
+  resources:
+  - '*'
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - generators.external-secrets.io
+  resources:
+  - acraccesstokens
+  - beyondtrustworkloadcredentialsdynamicsecrets
+  - cloudsmithaccesstokens
+  - ecrauthorizationtokens
+  - gcraccesstokens
+  - githubaccesstokens
+  - grafanas
+  - mfas
+  - passwords
+  - quayaccesstokens
+  - sshkeys
+  - stssessiontokens
+  - uuids
+  - vaultdynamicsecrets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - metallb.io
+  resources:
+  - bfdprofiles
+  - bgpadvertisements
+  - communities
+  - configurationstates
+  - ipaddresspools
+  - l2advertisements
+  - servicebgpstatuses
+  - servicel2statuses
+  verbs:
+  - get
+  - list
+  - watch
+
--- clusters/homelab Kustomization: flux-system/root ClusterRoleBinding: flux-system/mcp-kubernetes-readonly

+++ clusters/homelab Kustomization: flux-system/root ClusterRoleBinding: flux-system/mcp-kubernetes-readonly

@@ -0,0 +1,19 @@

+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  annotations:
+    kustomize.toolkit.fluxcd.io/prune: disabled
+  labels:
+    kustomize.toolkit.fluxcd.io/name: root
+    kustomize.toolkit.fluxcd.io/namespace: flux-system
+  name: mcp-kubernetes-readonly
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: mcp-kubernetes-readonly
+subjects:
+- kind: ServiceAccount
+  name: mcp-kubernetes-homelab
+  namespace: ai
+
`````

**Comment by ppat** [comment]
Thanks — this changed the design in two places, not just the wording. Closing the loop on each finding.

## Acted on

**1. `nodes/proxy` — dropped entirely, both clusters.** You're right and my documentation was wrong: I described it as reaching host log files, which understates a grant that authorises `GET /api/v1/nodes/<n>/proxy/exec/...`. It was the only rule in the role that wasn't read-only, and it failed the stated bar for this work ("read-only, nothing sensitive or abusable") on its own terms. `nodes_log` and `nodes_stats_summary` stop working; `nodes_top` is unaffected, which covers the node-resource-usage need that motivated the RBAC rework in the first place. The header comment now claims only the one remaining exception, `selfsubjectaccessreviews: create`.

This also moots finding 5 on ppat/homelab-ops-kubernetes-apps#3414 — the `disabled_tools` removal was only a gap because `nodes/proxy` was the one write channel RBAC couldn't catch. There is no longer such a channel.

**3. nas trimmed, and the asymmetry is now stated.** I'd anchored on "byte-identical across clusters" because it matches the existing `kubernetes-api-admin`/`readonly` convention; you're right that what the two clusters *hold* matters more than that symmetry. nas drops `longhorn.io` (it's NFS + MinIO), `tailscale.com`, `deviceplugin.intel.com` and `fpga.intel.com` — verified against `clusters/nas/kustomizations/`, which deploys no `*-extra` modules and no `infra-storage-core`.

`monitoring.coreos.com`, `dragonflydb.io` and the CNPG groups stay: nas is getting Prometheus exporters and promtail/alloy, and has CNPG + Dragonfly via Harbor.

`configmaps` and `pods/log` remain cluster-wide on nas, including the bitwarden namespace. That's a deliberate call, not an oversight — RBAC has no deny, so excluding a namespace means abandoning the ClusterRole for per-namespace Roles, and the operational cost of that isn't worth it here. Documented in-file so it stays a decision.

**5. `generators.external-secrets.io` — wildcard replaced with an explicit kind list.** You're right that this contradicted its own justification, and it's the same shape of argument I used to exclude `helm.cattle.io`, so consistency demanded acting. `fakes` and `webhooks` are excluded by name. `grafanas` stays granted with the trade-off now stated explicitly: it exposes a plaintext admin *username*, the password staying in an unreadable Secret.

Your cross-PR note about the `Grafana` generator being itself a counter-example was the thing that made this concrete rather than theoretical.

**6. Both residual-risk notes corrected.** You're right that `Middleware.spec.basicAuth` takes only a `secret` reference in the CRD — the inline surface is `spec.headers.customRequestHeaders` and `spec.plugin`. Tailscale's `ProxyClass` inline env (`TS_AUTHKEY`) added on homelab; nas drops the group entirely, so it doesn't apply there. These now live in the manifests rather than only in the PR body.

**7. `kube-system` → dedicated `mcp-access` namespace on nas.** Cheap, and stops this identity inheriting PSA/Kyverno exemptions added later.

## Deliberately deferred

**2. NetworkPolicy — deferred until Cilium.** The current CNI doesn't enforce NetworkPolicies, so writing one now would be inert. Cilium is already planned and brings enforcement with it.

I don't think that leaves the gap unmitigated in the meantime, because the most exposed bypass turned out to be one nobody had noticed: **n8n allowlisted every MCP Service directly** in `N8N_SSRF_ALLOWED_HOSTNAMES`, so its workflows could reach them without a LiteLLM key at all — routing around the virtual-key scoping that is the real access control, from the component most exposed to untrusted input. Those entries are removed in ppat/homelab-ops-kubernetes-apps#3414, so MCP traffic now has to traverse the gateway.

On your point that LiteLLM is claimed as the trust boundary without enforcement: MCP tool access *is* scoped per virtual key today, configured in the LiteLLM UI. That's a real control, but it's out-of-git and invisible to review, so it's now recorded in the module README along with the fact that registering a server doesn't grant existing keys access to it.

I also checked whether the server could enforce auth in-band instead. It can't, for this deployment: `require_oauth = true` and `cluster_auth_mode = "kubeconfig"` are mutually exclusive (`ValidateClusterAuthMode()` hard-errors), and with `require_oauth` on the server uses the *caller's* token as the cluster credential — which nas would reject. There's no static shared-secret mode and no TokenReview-based caller auth at any version through current `main`. A sidecar reverse-proxy would work; not building one now given Cilium's timeline, but it's the fallback if that slips. (Useful byproduct: `/healthz` is exempt from auth in every mode, so probes are safe under any future scheme.)

**4. Token rotation — deferred, revocation documented.** Building a bespoke rotation mechanism for one token isn't worth it ahead of a considered general approach. What's free and now in-file: revocation is `kubectl delete secret`, effective immediately, unlike a JWT with a baked-in expiry. Dropping `nodes/proxy` also shrinks this token's blast radius considerably — it is now genuinely read-only.

## One correction to the review

The `subPath` fix you proposed on ppat/homelab-ops-kubernetes-apps#3414 (`/etc/kubernetes-mcp-server/secrets`) would have kept the mount nested under the ConfigMap's own mount root and hit the same container-init failure — CI caught exactly that, `not a directory`, before the review landed. The diagnosis was right; the mount was moved to a sibling path instead.

Also worth recording: the fixture-drift risk you flagged materialised immediately — the CI copy was already behind the canonical role within the same session. Rather than rely on the sync comment, ppat/homelab-ops-kubernetes-apps#3414 now asserts `nodes/proxy`, `fakes` and `webhooks` are *denied*, so the specific regressions worth catching fail loudly instead of silently.

**Comment by ppat** [comment]
## Security review — round 2 (`ae75018`)

Re-reviewed the follow-up commit against the responses above. Short version: five of seven acted on, the two deferrals reasoned rather than waved away, and the header comment now describes the grant accurately rather than aspirationally. `nodes/proxy` gone from both clusters, the nas role genuinely narrowed with the exclusions justified against `clusters/nas/kustomizations/` rather than assumed, `mcp-access` replacing `kube-system`, and both residual-risk notes corrected to the right mechanism. The `configmaps`/`pods/log` cluster-wide decision on nas is now a documented decision instead of an omission, which is the part that matters.

One new finding, one gap created by an otherwise-correct fix, and a correction to my own finding 2.

---

### New: the `generators.external-secrets.io` allow-list still has two holes of the class it was built to close (Medium)

The wildcard → explicit-kinds change is right, and `fakes`/`webhooks` are the right two to exclude. But the list also grants `clustergenerators` and `generatorstates`, and both can carry the contents of the kinds just excluded.

**`ClusterGenerator` is a wrapper around any generator kind.** Verified against upstream `apis/generators/v1alpha1/types_cluster.go`: `ClusterGeneratorSpec` is `{kind, generator}`, where `generator` is a `GeneratorSpec` that inlines `FakeSpec` and `WebhookSpec` as fields. So:

```yaml
kind: ClusterGenerator
spec:
  kind: Fake
  generator:
    fakeSpec:
      data:
        password: hunter2      # identical plaintext, different envelope
```

`fakes` is denied, `clustergenerators` is allowed, the value is readable. This needs no adversarial creativity — the cluster-scoped form is simply what you reach for when a generator should be shared across namespaces.

**`GeneratorState` is ESO's own automatic copy.** From the v2.8.0 CRD bundle this repo pins, verbatim: `spec.resource` is *"the generator manifest that produced the state… a snapshot of the generator manifest at the time the state was produced"*, and `spec.state` is *"the state that was produced by the generator implementation."* So (a) a Fake's plaintext `spec.data` gets snapshotted into it, and (b) `spec.state` is generator **output** — which for `vaultdynamicsecrets`, which this role does grant, is lease material. These objects are written by the controller, not by a person, so they appear without anyone deciding to create one.

The root cause is worth stating once because it generalises: **RBAC allows and denies by kind, never by content.** Any kind that can embed another kind's content defeats a kind-level exclusion. That also means there's no safe way to keep `clustergenerators` — "allow ClusterGenerator unless `spec.kind` is Fake or Webhook" is precisely the field-level condition RBAC cannot express.

Fix is to drop both from the list on both clusters. Neither is in use today, so this is latent rather than live.

### The nas/homelab divergence is now the least-tested part of the role (Medium)

Narrowing nas was the right call, but it introduced a gap in the CI story: the fixture in ppat/homelab-ops-kubernetes-apps#3414 is a single `mcp-kubernetes-readonly` ClusterRole with both identities bound to it, and `check_identity` runs identical assertions against both. Production now has two different roles. A careless resync that widened nas back to homelab's shape would pass CI green — and re-adding `longhorn.io`/`tailscale.com`/the Intel groups is exactly the kind of edit a "make these two files consistent again" cleanup produces.

Agreed that a second chainsaw suite isn't the answer — this repo ships no modules, and standing up an integration suite for it would cost far more than it returns. It doesn't need one, because every invariant worth protecting here is a property of the YAML rather than of a running cluster:

- no verb outside `{get,list,watch}` anywhere, except `selfsubjectaccessreviews: create`
- `nodes/proxy` never appears
- core `secrets` never appears
- the `generators.external-secrets.io` rule contains no `*`, and none of `fakes`/`webhooks`/`clustergenerators`/`generatorstates`
- **nas ⊆ homelab**, minus a declared exclusion set

Two suggestions, in order of value:

1. **A subset assertion in pre-commit.** ~40 lines: parse both files with PyYAML, build the `{(apiGroup, resource, verb)}` set from each, assert `nas ⊆ homelab` and that the declared exclusion set (`longhorn.io`, `tailscale.com`, `deviceplugin.intel.com`, `fpga.intel.com`) is absent from nas. This is the only one of the invariants that can't be expressed per-file, and it's the one guarding the divergence just introduced.
2. **Kyverno CLI for the per-file invariants.** `kyverno apply` runs a `ClusterPolicy` against manifest files statically, no cluster needed, so one policy artifact covers both PR-time validation over `clusters/*/cluster/rbac/*.yaml` and — if deployed — admission-time enforcement that catches drift from any source, not just edits to this file. Reuses the policy language already in `policies/` rather than introducing Rego alongside it.

Conftest would also work, but adds a second policy language for no benefit I can see given Kyverno is already here.

### Correction to my finding 2 (NetworkPolicy)

The deferral premise turns out to be wrong — but so was half of my argument for doubting it, so both sides need amending.

The load-bearing fact is only this: **k3s has bundled the kube-router NetworkPolicy controller since 2019 and enables it unless `--disable-network-policy` is passed.** These clusters don't pass it, so policies are enforced today. That moves this from "deferred behind Cilium" back to actionable now, and the README line stating the CNI doesn't enforce them should come out.

My supporting argument — that the cluster already runs Flux and dragonfly-operator NetworkPolicies — was weak and should be struck. A policy shipped by a chart or a controller is just what that chart includes; with no enforcer it's inert YAML in etcd, and its presence proves nothing either way.

Operational note for whoever writes it, since these namespaces have never had an enforced policy: start with a targeted ingress policy on the MCP Services (podSelector on the `mcp-*` pods, ingress from the LiteLLM pod only) rather than a namespace-wide default-deny. Same benefit for this specific gap, much smaller blast radius. A default-deny in `ai` will also catch cross-namespace callers, API-server egress and metrics scraping that nobody is currently thinking about, and that debugging would delay the fix. Namespace-wide default-deny is the right end state, just not the right first commit.

### Corrections to my own review

- **The `subPath` fix path was wrong.** `/etc/kubernetes-mcp-server/secrets` would have hit the same container-init failure for the same reason — still nested beneath the ConfigMap volume's mount root. The sibling-path fix is correct. Diagnosis right, prescription wrong.
- **`defaultMode: 0400` was incomplete advice.** Without a pod-level `fsGroup` the file stays root-owned and a non-root container is denied outright; catching that mcp-grafana would have failed *silently* (lazy per-request token read behind a TCP-only probe) is a better catch than the original suggestion was a suggestion.

### Accepted, no further comment

- **Token rotation deferral** — reasonable, and stronger now: with `nodes/proxy` gone the token is genuinely read-only, and documented immediate revocation is the right minimum.
- **`configmaps`/`pods/log` cluster-wide on nas, bitwarden included** — the reasoning holds. RBAC has no deny, so carving out a namespace means abandoning the ClusterRole for per-namespace Roles, and that's a bigger change to this repo's RBAC model than the exposure warrants. Recording it in-file as a decision is the right resolution.
- **`monitoring.coreos.com` / `dragonflydb.io` / CNPG retained on nas** — justified; Harbor brings CNPG and Dragonfly.
- **Revised header comment** — accurate now. `selfsubjectaccessreviews: create` is correctly characterised as self-scoped, and the `nodes/proxy` absence documents the mechanism rather than restating the old understatement.

ppat:
## Round 2 response (`b9d4f80`)

Both new findings acted on. The generator-kind one was a real hole and the netpol correction changed the outcome rather than the wording.

### `clustergenerators` / `generatorstates` — dropped, both clusters

Correct, and I'd missed it in exactly the way the finding predicts: I checked which kinds *look* credential-bearing rather than which kinds can *carry* credential-bearing content. `ClusterGenerator.spec.generator` inlines `fakeSpec`/`webhookSpec`, so denying `fakes` while allowing its wrapper achieved nothing; `GeneratorState` is worse for being controller-written, so it appears without anyone choosing to create one, and `spec.state` is generator output — lease material for `vaultdynamicsecrets`, which this role does grant.

Worth noting the exclusion set is now internally consistent: the one genuinely output-bearing path through a granted kind is closed, rather than left as a documented caveat.

The generalisation is the durable part and it's now in `DESIGN.md` rather than only in this thread: **RBAC allows and denies by kind, never by content, so any kind able to embed another kind's content defeats a kind-level exclusion.** It also correctly forecloses the tempting alternative — "allow ClusterGenerator unless `spec.kind` is Fake" is a field-level condition RBAC cannot express.

### The divergence is now guarded — and building the guard found a bypass in itself

Agreed a second chainsaw suite would cost more than it returns, and agreed the invariants are properties of the YAML. Went with Kyverno CLI in GH Actions rather than pre-commit: pre-commit is bypassable with `--no-verify` and silently absent for anyone who hasn't installed the hooks, which is the wrong enforcement point for a security boundary. Policy lives in `ci/policy-tests/`, structurally unreachable by Flux (which only ever points at `./clusters/<name>` and `./policies/<group>`), so it cannot be deployed by accident.

**The part worth reporting: written in this repo's usual JMESPath style, the policy did not work.** Kyverno's JMESPath operators apply wildcard glob matching whenever either operand contains `*`. Because the role legitimately grants `resources: ["*"]` on one rule, that literal glob-matched unrelated targets — and an injected `verbs: ["*"]` **passed** the read-only check by wildcard-matching `get`/`list`/`watch`. A policy written to catch full-wildcard verbs would have waved through the single worst violation it exists to catch. Rewritten in CEL, where `in`/`==` compare literally.

That only surfaced because the policy was tested in the failing direction — each of `nodes/proxy`, core `secrets`, a non-read verb and a `fakes` grant injected in turn and confirmed to go red. It's the third control in this work that looked correct and wasn't, so that's now the standing expectation rather than a nicety.

`nas ⊆ homelab` is documented in the nas role's header as agreed, naming the homelab path to diff against, rather than approximated with a bespoke script.

### NetworkPolicy — you were right, the deferral premise was wrong

Confirmed and acted on rather than deferred. k3s bundles its own controller (a fork of kube-router's, imported via `replace`, not vendored) and enables it unless `--disable-network-policy` is passed, which these clusters do not pass.

I researched the probe question before writing anything, since it was the one failure mode that could take down all eight servers. **Probes are safe**: `pkg/controllers/netpol/pod.go` inserts an unconditional `-m addrtype --src-type LOCAL -j ACCEPT` at position 1 of every pod's firewall chain, ahead of any policy-chain jump — and k3s's fork leaves that file byte-identical to upstream. It's also a spec conformance requirement ("traffic to and from the node where a Pod is running is always allowed"), so it doesn't rest on one implementation.

Took your staging advice on scope — a targeted ingress policy on the MCP pods, not a namespace-wide default-deny — for the reason you gave: default-deny in `ai` would also catch cross-namespace callers, API-server egress and scraping nobody has enumerated. Details on ppat/homelab-ops-kubernetes-apps#3414.

One near-miss worth recording: the LiteLLM selector derived from the chart's `_helpers.tpl` is `app.kubernetes.io/name: litellm-helm`; the **live pod carries `litellm`**. Deriving the policy from chart source would have matched no source pod and denied the gateway outright — every MCP server dark behind a policy that reads as correct. Used the live labels, which also match the Service's own selector.

### Accepted as-is

Token rotation stays deferred with immediate `kubectl delete secret` revocation documented; `configmaps`/`pods/log` cluster-wide on nas including bitwarden remains a recorded decision; the retained groups on nas are justified. Your two self-corrections match what we found independently — the `subPath` path would have failed container init identically, and `defaultMode: 0400` without `fsGroup` breaks both pods.

PR #814, opened 2026-08-01
docs(cluster-homelab): add KubeVirt node-prep and Talos containerDisk runbooks

ppat:
## Summary

- Adds `OPERATIONS.md` (new file, no prior convention for host-level runbooks existed in this repo) with two manual runbooks that replace an unmaintained Packer/Ansible pipeline:
  - **Prepare a node for KubeVirt** — pre-flight gate for `/dev/kvm`, split into a "these two nodes, today" path (just the declarative label — both nodes already pass every check) and a "fresh or rebuilt node" path (the full sequence), the k3s `config.yaml.d` node-label change for `beelink-ser8-1`/`minisforum-nab9-1`, post-change verification, the Talos `machine.kernel.modules`/`machine.nodeLabels` equivalent, and rollback.
  - **Build and publish the Talos containerDisk** — a `crane`-based (no daemon needed) script that resolves a Talos Image Factory schematic ID, downloads `nocloud-amd64.raw.xz`, converts it to sparse qcow2, wraps it as a `FROM scratch` OCI image with the disk at `/disk/` (owned `107:107`), and pushes it to Harbor.
- Cross-links `OPERATIONS.md` from `README.md`'s "Finding your way" table and `CLAUDE.md`'s "Start here" list, matching this repo's existing doc cross-linking convention.

## Fixed: a real DNS domain was leaking into the node label (this update)

Every command in this runbook (the `kubectl label` in Path A, the Path B `config.yaml.d` drop-in, post-change verification, the Talos `nodeLabels` equivalent, and rollback) used the operator's real DNS domain (redacted here) hardcoded into what should have been a module-generic parameter. That mirrored the same bug in `kustomizations/infra-virtualization-core.yaml` (#817), which exposed the label as two Flux `postBuild` variables (`vm_node_label_key`/`vm_node_label_value`) that a consuming cluster could set to anything, including its own real domain — the label key is the KubeVirt module's own scheduling contract, not cluster-specific config.

Fixed at the source: apps repo PR [ppat/homelab-ops-kubernetes-apps#3506](https://github.com/ppat/homelab-ops-kubernetes-apps/pull/3506) removes both substitution variables and hardcodes the label as a module-owned constant, `homelab-ops.internal/virtualization: "enabled"` — `.internal` is a reserved, non-delegatable TLD (RFC 9476), chosen because this repo has no established convention of its own for self-defined label prefixes. Every occurrence in this runbook now uses that constant, and a new sentence up front states plainly that it's a fixed constant, not a per-cluster choice, so a future edit doesn't reintroduce a real domain here by habit.

**Action needed if you already ran the old `kubectl label` command from this runbook against `beelink-ser8-1`/`minisforum-nab9-1`:** apply the new label and remove the old one —

```bash
kubectl label node beelink-ser8-1 homelab-ops.internal/virtualization=enabled <your-old-real-domain>/virtualization-
kubectl label node minisforum-nab9-1 homelab-ops.internal/virtualization=enabled <your-old-real-domain>/virtualization-
```

If you haven't labelled either node yet, just use the (now-corrected) command in the runbook below directly.

## Review follow-ups (previous update)

An adversarial review of the first version of this PR found four issues, all addressed:

1. **`node-label:` → `node-label+:`.** Verified against the current [k3s config docs](https://docs.k3s.io/installation/configuration): `config.yaml.d` files merge by key, last-file-wins, *unless* the key carries a `+` suffix, which appends instead. A bare `node-label:` would let any future drop-in that also sets `node-label:` silently erase this one — exactly the disaster-recovery re-registration scenario the drop-in exists to serve.
2. **The pre-flight gate now actually gates.** It computed pass/fail and printed a warning but never exited non-zero. It now `exit 1`s on any FAIL (and `exit 0` on pass), so it can gate a wrapper script instead of scrolling past.
3. **Split "what a fresh node needs" from "what these two nodes need now."** The operator has since run the pre-flight against both `beelink-ser8-1` and `minisforum-nab9-1` — everything KubeVirt needs is already present. That run also surfaced that `kvm_intel`/`kvm_amd` already autoload via the kernel's CPU-feature match (`MODULE_DEVICE_TABLE(x86cpu, ...)`, VMX/SVM) and `tun`/`vhost_net` autoload on first device open via udev's devname alias — with no `modules-load.d` config on either node. The runbook's original kernel-module persistence step is dropped entirely: it added nothing already-true, and unconditionally listing both `kvm_intel` and `kvm_amd` would have guaranteed `systemd-modules-load.service` fails on every boot on both nodes (no CPU has both VMX and SVM). The pre-flight script itself had a related false-negative — it gated `tun`/`vhost_net` on `lsmod`, which won't show either module until something first opens the device node — fixed to check device-node existence instead. Runbook now has two clearly labeled paths (today vs. fresh/rebuilt node) so a reader doesn't drain a node for no reason.
4. **The `msr`/Secure-Guest `virt-host-validate` `WARN`s are confirmed non-blocking and out of scope for KubeVirt** — `/dev/cpu/0/msr` is unrelated CPU-introspection tooling, and Secure-Guest support (SEV/TDX) isn't used by either planned VM. Neither is part of the gate; both are expected on this hardware. IOMMU-enabled status is also now recorded in the runbook (relevant for future PCI/GPU passthrough).

The `kubectl drain`-before-stopping-k3s suggestion from the same review was **not** applied — the operator already has a faster, trusted cordon-without-drain mechanism via `system-upgrade-controller` for the same class of operation, and this runbook keeps to it rather than introducing a second convention.

## Merge order

This PR and #815 (`feat/kubevirt-sandbox-cages`) both add `OPERATIONS.md`, which doesn't exist on `main`. Whichever merges second will hit a small, purely-additive conflict on that file plus the one-line link additions in `CLAUDE.md`/`README.md`. **Recommend merging this PR (#814) first** — its KubeVirt node-prep runbook is the prerequisite context for the before/after baseline run that #815's netpol-falsifiability-probe section of `OPERATIONS.md` depends on. Not resolved pre-emptively here to keep the two PRs' changes independent.

## Test plan

- [x] `markdownlint-cli2 --config .markdownlint-cli2.yaml OPERATIONS.md README.md CLAUDE.md` — passes
- [x] `shellcheck` on every embedded `bash` script block — passes
- [x] `pre-commit run --all-files` — passes
- [x] Manually ran the pre-flight script's logic against this workspace's sandbox to confirm it exits 0 on all-PASS and 1 on any FAIL
- [ ] Manual: run the pre-flight script on `beelink-ser8-1` and `minisforum-nab9-1` (already done by the operator per the review round — informed this update) and confirm the `homelab-ops.internal/virtualization=enabled` label lands correctly (relabel per the "Action needed" note above if the old real-domain-based label was already applied)
- [ ] Manual: dry-run the containerDisk build script end-to-end (crane push auth, actual qcow2 size) on first real use

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,6 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - [DESIGN.md](./DESIGN.md) — directory anatomy, how a module gets wired into a cluster (sources/kustomizations/dependsOn/components/postBuild/patches), the `services/` pattern, secrets/RBAC/storage model, CI, versioning
 - [clusters/homelab/README.md](./clusters/homelab/README.md) / [clusters/nas/README.md](./clusters/nas/README.md) — what's actually deployed on each cluster, with links to each module's own README in the apps repo
 - [policies/README.md](./policies/README.md) — Kyverno policy groups and what's enforced where
+- [OPERATIONS.md](./OPERATIONS.md) — manual, host-level runbooks (node prep, image builds) for procedures that happen outside Flux
 
 This is a GitOps repo (FluxCD) of Kustomize manifests — there is no application
 code to build. "Development" here means authoring/editing YAML and validating
--- a/OPERATIONS.md
+++ b/OPERATIONS.md
@@ -22,6 +22,12 @@ runbook verifies a node is actually capable and gives it a declarative label
 so KubeVirt VMs can be scheduled onto it deliberately (`nodeSelector`), not
 just wherever a node happens to be capable.
 
+The label key below, `homelab-ops.internal/virtualization`, is a fixed
+constant the `infra-virtualization-core` module itself hardcodes (see that
+module's README in the apps repo) — it's this cluster's scheduling contract
+with the module, not a value chosen per-cluster, so every command in this
+runbook uses it verbatim.
+
 **Target nodes:**
 
 | Node | IP | CPU | Memory | Why it qualifies |
@@ -192,7 +198,7 @@ ssh beelink-ser8-1 'sudo install -o root -g root -m 0644 -D /tmp/90-kubevirt.yam
 # 2. Apply the label directly -- this is the part that actually takes effect
 #    today, since k3s only reads config.yaml.d node-label(+) at registration
 #    time and both nodes are already registered.
-kubectl label node beelink-ser8-1 homelab.nikara.net/virtualization=enabled
+kubectl label node beelink-ser8-1 homelab-ops.internal/virtualization=enabled
 
 # Repeat both steps for minisforum-nab9-1.
 ```
@@ -235,7 +241,7 @@ third node added to the KubeVirt pool later.
    # error -- precisely the disaster-recovery re-registration scenario this
    # file exists to serve.
    node-label+:
-     - "homelab.nikara.net/virtualization=enabled"
+     - "homelab-ops.internal/virtualization=enabled"
    ```
 
 3. Let the node join normally. Registration applies the label automatically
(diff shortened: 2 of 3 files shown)


ppat:
Part of the KubeVirt sandbox design in ppat/homelab-ops-kubernetes-apps#3503 (Phase 1 — node enablement runbook). Status and links to the rest of the implementation are tracked there.

PR #815, opened 2026-08-01
feat(cluster-homelab): add KubeVirt sandbox namespace isolation (Phase 3)

ppat:
## Summary

Phase 3 of the KubeVirt onboarding plan: builds the isolation "cages" for
two planned VMs — a dockerd VM and a single-node Talos sandbox AI agents
get full write access to — before either VM exists.

- Two new namespaces under `clusters/homelab/services/`: `sandbox-docker`,
  `sandbox-talos`.
- Default-deny **ingress and egress** NetworkPolicies in both (the first
  egress-enforcing policies authored directly in this repo — every existing
  NetworkPolicy in the apps repo is ingress-only).
- Egress: DNS to CoreDNS only, same-namespace, and `0.0.0.0/0` except
  `10.0.0.0/8` / `172.16.0.0/12` / `192.168.0.0/16` / `169.254.0.0/16` /
  `127.0.0.0/8` — that `except` list is what closes the prod
  kube-apiserver, Longhorn's manager API, kubelet, etcd, the UniFi gateway,
  and the NAS.
- Ingress differs by design: `sandbox-docker` admits only the `coder`
  namespace (SSH + a documented `kind` `apiServerPort` range);
  `sandbox-talos` admits `coder` and the `ai` namespace's `mcp-server` pods
  (mirrors `apps-ai`'s existing `mcp-servers-ingress` selector shape), never
  `sandbox-docker` — a reachable dockerd is root-equivalent on that VM, so
  the two sandboxes must not reach each other.
- Name-scoped (never namespace-wide) Kyverno exclusions for the Baseline and
  Restricted policies KubeVirt's components will actually trip — see
  "Review follow-ups" below for what changed here.
- A `netpol-falsifiability-probe` CronJob per namespace that asserts every
  denial still holds *and* that internet egress + CoreDNS resolution still
  work, failing the Job if a denial silently stops enforcing (queryable via
  `kube_job_failed`; no AlertManager route by design). `OPERATIONS.md`
  documents the before/after baseline run that makes this probe falsifiable
  rather than decorative.

No VM exists yet — Phase 4/5 drop the two VMs into these namespaces on top
of this.

## Review follow-ups (this update)

The review correctly flagged that this repo's `homelab` cluster runs
Kyverno's **Restricted** PSS tier (adds `disallow-capabilities-strict`,
`disallow-privilege-escalation`, `require-run-as-non-root-user`,
`require-run-as-nonroot`, `restrict-seccomp-strict`, `restrict-volume-types`
on top of Baseline), and the original exclusions here only covered the 4
Baseline policies the brief named literally, declining to guess at the
Restricted gap without direction.

Verified per-component against upstream KubeVirt's actual generated
manifests (Deployment/DaemonSet Go source, not assumptions or OpenShift SCC
docs describing a different ceiling):

- **`virt-handler`** (DaemonSet) is genuinely privileged, `hostPID`, runs
  root, declares no seccomp profile, and hostPath-mounts libvirt/kubelet/
  node-labeller state. It trips all 6 Restricted policies, plus (as
  originally scoped) 3 of the 4 Baseline ones — see below on the 4th.
- **`virt-launcher`** (the per-VM pod) is *not* privileged: KubeVirt's
  default non-root mode drops all capabilities except `NET_BIND_SERVICE`,
  sets `runAsNonRoot`/`runAsUser=107`, `allowPrivilegeEscalation=false`,
  gets `/dev/kvm`/`tun`/`vhost-net` via device-plugin resource requests
  rather than a hostPath volume, and uses only allow-listed volume types.
  The one real gap: it never declares a `seccompProfile` (the KubeVirt CR's
  `seccompConfiguration` isn't set in this repo), so it trips exactly one
  Restricted policy — `restrict-seccomp-strict` — and nothing else.
- **`virt-operator`/`virt-api`/`virt-controller`** are ordinary Deployments:
  non-root, drop-all capabilities, no hostPath, no host namespaces. They
  trip nothing in either tier and get no exclusions anywhere.

That same verification also caught that the already-merged Baseline
exclusions were over-scoped: `virt-launcher*`/`virt-operator*` were excluded
from `disallow-capabilities`/`disallow-host-path`/`disallow-privileged-containers`
even though neither ever trips them, and `disallow-host-namespaces` excluded
all three components from its `hostNetwork` and `hostIPC` rules when
`virt-handler` only ever sets `hostPID`. Narrowed all four Baseline files to
`virt-handler*` only, on exactly the rules it actually trips — excluding a
component from a policy it never violates is a standing false grant the
moment that component's manifest changes, not merely redundant today. Also
narrowed every namespace list from `[kubevirt, sandbox-docker, sandbox-talos]`
to only where each component can actually be named (`virt-handler` only ever
runs in `kubevirt`; `virt-launcher` only in the two sandbox namespaces).

Every exclusion is name-scoped, never namespace-wide (the existing
`intel-gpu-plugin-gpu-device*` precedent), and its comment now states three
things: what's tolerated in plain terms (not "excludes X from Y" — the diff
already says that), why KubeVirt genuinely needs it tied to the actual
mechanism, and what bounds the risk. For `virt-handler`'s privilege
specifically: once KubeVirt is installed its CR's
`spec.workloads.nodePlacement.nodeSelector` restricts it to the two nodes
`OPERATIONS.md`'s KubeVirt runbook prepares (#814), not all four, and
`virt-launcher` — where a VM's guest actually executes — is unprivileged.
The residual this doesn't close is recorded rather than left implicit: a
guest escape reaching the host needs a second bug at exactly the
`virt-handler`↔`virt-launcher` boundary, which has a live CVE lineage
(CVE-2025-64437, CVE-2025-64433, both fixed in 1.5.3/1.6.1) — pin to a
patched KubeVirt version.

All policies stay `validationFailureAction: Audit`; nothing is blocked
today. These exclusions are written to the shape Enforce would need so that
moving to Enforce later doesn't require re-deriving any of this.

## Documented: why kubevirt-namespace ingress stays blocked (this update)

A source-level verification of KubeVirt v1.9 established that every
operational path — VM lifecycle (start/stop/pause/exec, domain stats) over a
node-local unix socket on a shared hostPath, console/VNC/serial and metrics
via `virt-handler`'s own pod IP inside the `kubevirt` namespace — is
node-local or stays within `virt-handler`, and never needs a route into
`sandbox-docker`/`sandbox-talos`. The one exception is `virtctl ssh`/
`virtctl port-forward`: `pkg/virt-api/rest/portforward.go` uses `netDial`
rather than `virtHandlerDialer`, so `virt-api` dials the VM's masquerade pod
IP directly — that *does* cross into the sandbox namespace.

**Decision: leave it blocked, don't add a `kubevirt`-namespace ingress
rule.** Nothing operational depends on it; `kubectl port-forward` is a fully
equivalent, unaffected substitute (containerd's CRI dials `localhost` inside
the pod's own network namespace, never reaching the veth NetworkPolicy
enforces on); and a namespace-scoped admit would also let in `virt-api`/
`virt-operator`/`virt-controller` alongside the already-privileged
`virt-handler` — not a cost worth paying for a path with a working
substitute.

That decision is otherwise invisible (it manifests as nothing being there),
so it's recorded in both places a future reader would look:

- A comment next to each namespace's ingress `NetworkPolicy`
  (`allow-coder-ingress` in `sandbox-docker`, `allow-coder-and-mcp-ingress`
  in `sandbox-talos`) stating the absence is deliberate, why it's safe, the
  supported alternative, and what admitting `kubevirt` would cost.
- A new `OPERATIONS.md` runbook, **"reach the sandbox VMs"**, with the
  actual `kubectl port-forward` commands for the Talos API (`:50000`), the
  sandbox Kubernetes API (`:6443`), and the Docker VM's SSH (`:22`), each
  pointing the client at `127.0.0.1` — so `virtctl ssh`'s failure is
  pre-explained, not debugged. Also notes Talos ships no SSH daemon at all
  (API-only via `talosctl` on `:50000` with mTLS), so the SSH question
  doesn't arise for that VM regardless of NetworkPolicy.

## Merge order

This PR, #814 (`docs/kubevirt-talos-node-runbooks`), and #817
(`feat/wire-infra-virtualization-core`) all independently added
`OPERATIONS.md`, which didn't exist on `main` when any of the three were
opened. **#814 has since merged** (`OPERATIONS.md` now carries its
node-prep and Talos containerDisk runbooks on `main`), so this branch has
been rebased onto current `main` and its `OPERATIONS.md` conflict resolved
here: the "verify the sandbox NetworkPolicy falsifiability probes" and
"reach the sandbox VMs" runbooks now sit after #814's two runbooks in one
file, and the `CLAUDE.md`/`README.md` one-line pointers to it are merged
rather than duplicated. #817 still creates `OPERATIONS.md` independently
and will need the same rebase-and-fold treatment on its own branch before
it can merge.

## Test plan

- [x] `pre-commit run --all-files` passes (yamllint, markdownlint,
      shellcheck, kubeconform-based manifest validation).
- [x] `kustomize build clusters/homelab` and each new
      `services/sandbox-{docker,talos}/` directory build cleanly.
- [x] `kustomize build` on `policies/pod-security-standard/baseline` and
      `.../restricted` build cleanly with the new exclusions.
- [x] Embedded probe shell script extracted and run through `shellcheck -s
      sh` clean.
- [ ] Not applied to any live cluster (no write access; CI/operator only).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - [DESIGN.md](./DESIGN.md) — directory anatomy, how a module gets wired into a cluster (sources/kustomizations/dependsOn/components/postBuild/patches), the `services/` pattern, secrets/RBAC/storage model, CI, versioning
 - [clusters/homelab/README.md](./clusters/homelab/README.md) / [clusters/nas/README.md](./clusters/nas/README.md) — what's actually deployed on each cluster, with links to each module's own README in the apps repo
 - [policies/README.md](./policies/README.md) — Kyverno policy groups and what's enforced where
-- [OPERATIONS.md](./OPERATIONS.md) — manual, host-level runbooks (node prep, image builds) for procedures that happen outside Flux
+- [OPERATIONS.md](./OPERATIONS.md) — manual, occasional runbooks for procedures that happen outside Flux (node prep, image builds, sandbox policy verification/access)
 
 This is a GitOps repo (FluxCD) of Kustomize manifests — there is no application
 code to build. "Development" here means authoring/editing YAML and validating
--- a/OPERATIONS.md
+++ b/OPERATIONS.md
@@ -1,15 +1,17 @@
 # Operations
 
-This document holds manual, host-level runbooks: procedures that happen
-*outside* Flux, against a physical node or an external registry, because they
-can't be expressed as a reconciled manifest. Everything else in this repo is
-declarative and covered by [DESIGN.md](./DESIGN.md) instead — start here only
-for the operations listed below.
+This document holds manual, occasional runbooks: procedures that happen
+*outside* Flux — against a physical node, an external registry, or as an
+on-demand check of a reconciled resource's actual runtime behavior — because
+they can't be expressed as, or fully verified by, a reconciled manifest
+alone. Everything else in this repo is declarative and covered by
+[DESIGN.md](./DESIGN.md) instead — start here only for the operations listed
+below.
 
-These runbooks used to be automated by a Packer/Ansible pipeline that is now
-unmaintained and being archived; until the planned k3s → Talos migration
-removes the need for host-level node prep entirely, they're accepted as
-manual, occasional, human-run procedures instead.
+The host-level runbooks below used to be automated by a Packer/Ansible
+pipeline that is now unmaintained and being archived; until the planned k3s →
+Talos migration removes the need for host-level node prep entirely, they're
+accepted as manual, occasional, human-run procedures instead.
 
 ## Runbook: prepare a node for KubeVirt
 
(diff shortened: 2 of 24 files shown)


ppat:
Part of ppat/homelab-ops-kubernetes-apps#3503 (Phase 3 — sandbox namespace isolation, built before either VM exists). Status and links to the rest of the implementation are tracked there.

PR #818, opened 2026-08-02
fix(cluster-homelab): make the sandbox netpol falsifiability probe self-establish enforcement

ppat:
## Summary

The `netpol-falsifiability-probe` in `sandbox-docker`/`sandbox-talos` reported **total failure** on its first real run — every deny target (prod kube-apiserver, kubelet, etcd, the LAN gateway, the in-cluster `kubernetes` Service) came back reachable. That triggered a long investigation which eliminated ipsets, `ipBlock.except`, rule ordering, egress support, and cluster-wide NetworkPolicy failure, in turn, before landing on the real cause:

**The probe pod itself was never policed.** k3s's bundled kube-router-derived NetworkPolicy controller programs each pod's firewall chain and its `KUBE-ROUTER-FORWARD` dispatch rule *reactively*, after the pod exists. A short-lived `Job` pod can complete before that rule is ever programmed — falling straight through to `-P FORWARD ACCEPT`, fully unpoliced. **The NetworkPolicies are correct** (proven by hand against a long-lived pod: `nc -zv -w3 192.168.8.1 443` → blocked); **the test was structurally incapable of testing them.**

## What changed

- **Vehicle: CronJob/Job → long-lived Deployment.** A long-lived pod is programmed once and stays programmed, rather than re-running the same creation race every 15 minutes. A bounded warm-up gate (poll a known-deny target until denial is actually observed, `WARMUP_TIMEOUT_SECONDS=30`) runs before every probe cycle and fails loudly — distinct log line, non-zero exit, pod restart/`CrashLoopBackOff` — if enforcement was never established for *this* pod, instead of silently trusting it. That gate re-runs on every container (re)start (rollout, node drain, eviction, crash), so the creation race stays exercised, just less often than every 15 minutes.
- **Dropped the Longhorn-manager (`:9500`) deny target.** Verified against the chart: Longhorn's manager Service is ClusterIP-only, no `hostNetwork`/`hostPort`/`NodePort`. `node-IP:9500` was a dead port regardless of any policy and reported a false "denied as expected" on every run — a deny-check against a target nothing listens on passes forever and proves nothing. `kubelet` (`:10250`, genuinely `hostNetwork` on every node) already covers "does the deny mechanism work against every node IP".
- **Reworded verdict lines.** `FAIL: prod kube-apiserver (...) -- connected, expected denial` reads, at a skim, like the opposite of what happened — that ambiguity is exactly what slowed down root-causing this. Now: observation first, verdict after — `REACHABLE (violation)` / `BLOCKED (ok)`.
- **Per-cycle `SUMMARY` line** so "the probe died" and "the probe is cleanly passing" stay distinguishable in Loki — an absent recent `SUMMARY` is itself a signal, not silence.

## Where the caveat is documented

The pod-creation enforcement window is a **real, standing security property**, not just a testing artifact — an agent with pod-create rights in `sandbox-talos` could run a Job that exfiltrates in its first seconds and exits before enforcement arrives. Documented in two places:

- A `CAVEAT` comment on the `default-deny` policy in both `sandbox-docker/network-policy.yaml` and `sandbox-talos/network-policy.yaml`.
- A new **"Known limitation: the per-pod NetworkPolicy enforcement window"** section in `OPERATIONS.md`, covering what the window is, why it exists, what it does/doesn't affect (the sandbox VMs are unaffected — long-lived, policed within seconds), and how to independently verify a given pod is actually policed: `iptables-legacy -S KUBE-ROUTER-FORWARD | grep <pod-ip>` from a `kubectl debug node/<node> --profile=sysadmin` pod, checked against **the node the pod actually landed on**, using `apk add iptables-legacy` since k3s no longer bundles `iptables` and the rules live in the legacy tables (invisible to `nft list ruleset`).

The existing before/after verification runbook in `OPERATIONS.md` is also rewritten for the Deployment shape (baseline now needs an ad hoc pod since the warm-up gate can't succeed with no policy in place — that crash-loop is the expected baseline signature, not a bug), and adds the LogQL queries for the probe's ongoing state now that there's no `kube_job_failed` metric.

## Verification

- `kustomize build clusters/homelab` — clean.
- `pre-commit run --all-files` — green (yamllint, shellcheck, markdownlint, kubeconform via `validate-k8s`).
- Confirmed against the live `loki` datasource that the probe's per-line verdicts are already captured with `namespace`/`pod`/`node_name`/`container` labels, independent of Job vs. Deployment — this is what OPERATIONS.md's new LogQL queries rely on.
- **Not verified against a live cluster** (no cluster access from this environment): the actual warm-up timing (`WARMUP_TIMEOUT_SECONDS=30`, `WARMUP_POLL_INTERVAL_SECONDS=2`) against k3s's real kube-router sync period — chosen generously but not re-derived from source for this change. Worth confirming on first real rollout that warm-up consistently completes in low single-digit seconds, not tens of seconds.

## Housekeeping

**`probe-long`** — a long-lived pod created manually in `sandbox-docker` during the investigation (used to prove the NetworkPolicies work correctly once a pod is warmed up) — may still be running. It is unrelated to this change and was deliberately left alone; needs manual `kubectl delete pod probe-long -n sandbox-docker` cleanup.

## Test plan

- [ ] Land this PR, then follow the rewritten "Runbook: verify the sandbox NetworkPolicy falsifiability probes" in `OPERATIONS.md` against the real cluster (baseline → apply policy → confirm the flip) for both namespaces.
- [ ] Confirm `WARMUP-OK` timing in practice is comfortably under `WARMUP_TIMEOUT_SECONDS`.
- [ ] Manually clean up `probe-long` in `sandbox-docker`.


ppat:
Part of ppat/homelab-ops-kubernetes-apps#3503 — fixes the netpol falsifiability probe added in #815 (Phase 3). The root cause (k3s's per-pod firewall-chain enforcement race) and the fact the NetworkPolicies themselves proved correct once the probe was fixed are both recorded on #3503.

PR #825, opened 2026-08-02
fix(cluster-homelab): fix two independent breaks in the Talos containerDisk runbook

ppat:
## Summary

#823 landed a working Talos containerDisk runbook. Two changes that landed independently since then jointly break it — neither wrong on its own, the same class of failure as pinning a value next to something else that manages it. Both are fixed here, and this is blocking the whole Talos sandbox phase (the VM module in `homelab-ops-kubernetes-experiments` is CI-green but can't deploy without a containerDisk to reference).

### 1. `apt-get install qemu-utils` on the Docker VM's host OS now fails outright

`homelab-ops-experiments` #209 (slimming that VM's root filesystem, landed independently) deletes `/var/lib/apt/lists` on every boot and masks the `apt-daily` timers that would otherwise repopulate it. That VM's root is itself a containerDisk that resets to its pristine image on every restart anyway, so #823's "leave `qemu-utils` installed for next time" advice was wrong from the start, not just newly broken — fixed here too.

**Fix:** run `qemu-img` in a throwaway container instead. #823 already tried this and hit an indefinite hang, diagnosed (but never actually tested) as Docker's default seccomp profile blocking `io_uring_setup`. Testing that diagnosis directly this time: `--security-opt seccomp=unconfined` alone does **not** fix the hang — reproduced at both a trivial 10MiB file and the full 4.15GiB Talos image, for both `qemu-img info` and `qemu-img convert`. The actual blocker is Docker's default **AppArmor** profile: `--security-opt apparmor=unconfined` alone (seccomp left at its normal default) fixes it immediately, verified at both scales and cross-checked `sha256sum`-identical against the original host-OS conversion's output. The runbook states the AppArmor relaxation and its justification explicitly (throwaway container, offline conversion, network-isolated VM) rather than passing over it.

### 2. Even fixed, there's nowhere safe on the VM to stage the 4.4GB of scratch data

- **Root filesystem**: 2.3GB total, as little as 109MiB free (measured live). Doesn't fit regardless of how it's spent.
- **`/tmp`**: a 7.9GB tmpfs — would fit, and an earlier version of this pipeline may well have "worked" there in testing, which is exactly the trap: it's RAM-backed, this cluster's VMs are not memory-overcommitted, and staging 4.4GB there silently burns a quarter of the guest's real memory reservation for the duration of the build.
- **A plain Docker named volume**: not a safe default either. This VM has a separate, pre-existing bug where containerd's own snapshotter writes image/layer data to `/var/lib/containerd` on the *root* disk instead of the PVC-backed `/var/lib/docker` that `docker info` reports as its data root (confirmed live: 212MiB stranded on root from that alone). Trusting a named volume lands correctly right now is exactly the kind of untested assumption this runbook exists to stop making.

**Fix:** bind-mount an explicit, PVC-backed host directory into the container as its working directory — the same pattern the `docker-host` module already uses for its own install/cache/log directories (a neutral top-level path bind-mounted from a `/var/lib/docker` subdirectory, since `dockerd` chmods `/var/lib/docker` itself to `0710` on every start and blocks direct traversal). That module is being extended with a general-purpose scratch directory in parallel with this fix, but hadn't landed as of this PR — rather than invent a path and risk the two drifting silently, the script takes it as a required `DOCKER_VM_SCRATCH_DIR` input with **no default**. Documented assumption: `/opt/docker-scratch`, matching that module's existing `/opt/docker-install` naming — **confirm the actual path against the docker-host module before running this**.

## What was proven by execution vs. inspection

Both fixes were proven by actually extracting the finished script from `OPERATIONS.md` and running it end to end against a real Talos v1.13.7 Factory asset (empty schematic):

- Downloaded and decompressed the real Factory image.
- Converted it to qcow2 in a throwaway `alpine` container, bind-mounted to a temporary PVC-backed stand-in directory (built the same way the `docker-host` module builds its own bind mounts, torn down afterward).
- Verified `sha256sum`-identical qcow2 output across **three independent conversion methods**: host OS directly, container + named volume, container + PVC bind-mount.
- Built the containerDisk OCI image with `crane` and inspected it directly (single scratch layer, `disk/talos.qcow2`, owned `107:107`) rather than trusting it.
- Watched disk usage during the real run: `/var/lib/docker` (PVC) rose from 449MiB to 1.1GiB; the VM's root filesystem moved by only ~23MiB of incidental overhead — confirming the scratch data actually lands where it's supposed to, not just in theory.

Not done, deliberately: no push to Harbor (Phase 4 stays the operator's step — no Tailscale connectivity or credentials in this environment), nothing applied to any cluster.

## Test plan

- [x] `pre-commit run --all-files` passes
- [x] Extracted the script verbatim from `OPERATIONS.md`, `bash -n` and `shellcheck` clean
- [x] Ran the extracted script end to end against a real Talos v1.13.7 asset via a temporary PVC-backed scratch directory, produced a valid containerDisk image
- [x] Verified qcow2 output is byte-identical across host-OS, container+volume, and container+PVC-bind-mount conversion methods
- [x] Docker VM left exactly as found: temporary verification bind-mounts unmounted and removed, containers removed, root/PVC disk usage back to baseline
- [ ] Operator: confirm the actual `DOCKER_VM_SCRATCH_DIR` path once the `docker-host` module's scratch-directory addition lands, and run Phase 4 (`crane push`) from a machine with Harbor credentials and Tailscale connectivity

### DIFF
--- a/OPERATIONS.md
+++ b/OPERATIONS.md
@@ -420,6 +420,18 @@ explicitly rejected in favor of building from Talos's own official artifact:
 it's more work, but it avoids depending on one person's unmaintained image
 and is a more realistic rehearsal of the eventual Talos migration.
 
+**Prerequisite — check before running this, don't discover it mid-run:**
+this runbook stages the Factory image in a PVC-backed scratch directory on
+the Docker VM (`/opt/build-scratch` — see "Where the ~4.4GB of scratch data
+has to live" below for why). That directory is created by the `docker-host`
+module in `homelab-ops-kubernetes-experiments`, as of this writing still
+open as experiments PR #212, not yet merged. This runbook cannot succeed
+until that PR is merged *and* the Docker VM has been restarted — the
+directory is created by cloud-init, which only runs at boot, so merging
+alone isn't enough. If `/opt/build-scratch` doesn't exist yet, that's why;
+don't route around it (see the script's own comments on why it fails
+instead of falling back to `sudo`).
+
 ### Choosing the build tool: `crane`, no daemon required
 
 KubeVirt's `containerDisk` requirement (verified against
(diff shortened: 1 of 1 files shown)


ppat:
Pushed a follow-up commit addressing review feedback:

1. **Scratch path corrected**: `/opt/build-scratch` (as landed in `homelab-ops-kubernetes-experiments` PR #212, still open), not `/opt/docker-scratch`. Added an explicit prerequisite note at the top of the runbook — this can't succeed until #212 merges *and* the Docker VM restarts (cloud-init only runs at boot). Also documented #212's deliberate root-owned mountpoint-stub design and why the script must not paper over the resulting `Permission denied` with `sudo`/`mkdir -p` — that failure is the intended signal, not a bug.

2. **AppArmor finding sharpened**: both the runbook prose and the script's own header comment now say plainly and prominently that seccomp is *not* the cause and was tested directly (twice, at two file sizes) before being ruled out, plus the mechanism (AppArmor mediating a mount-related operation when QEMU opens an existing file) and a cross-reference to `ppat/coder#846` — an independently diagnosed failure with the same root cause.

3. **Fixed a repeat bash bug**: an apostrophe inside a `${VAR:?message}` block broke script parsing again, the same way it did earlier in this PR. Re-verified with `bash -n`/`shellcheck`, and re-ran the full script end to end against a real Talos v1.13.7 asset through a temporary PVC-backed scratch directory — output `sha256sum`-identical to every prior conversion in this investigation.

Also removed a stray exited container and cached `alpine:3.20` image left on the Docker VM from earlier verification runs — my prior "left exactly as found" claim was ahead of the facts. Re-verified this time: no containers, no images beyond the pre-existing `hello-world`, no volumes, no stray mounts, root/PVC disk usage at baseline.

`pre-commit run --all-files` green.

PR #837, opened 2026-08-03
fix(cluster-homelab,policies): scope kubevirt read into mcp-kubernetes-readonly and close the controllerrevisions bypass

### LINKED ISSUE #824: Research: is it safe to add kubevirt.io read access to the mcp-kubernetes-readonly role?
## Status

**Phase 6 scope**, decided alongside the write-capable Kubernetes MCP work this research feeds directly into — that MCP is the reason this question needs answering at all, and it's being built in the same phase. See ppat/homelab-ops-kubernetes-apps#3503 for the full Phase 6 breakdown, including why the scope was drawn where it was.

## Why this came up

KubeVirt is now deployed and hosts two sandbox VMs. AI agents reach Kubernetes solely through the MCP server, and its `mcp-kubernetes-readonly` ClusterRole does **not** include `kubevirt.io`. So agents cannot see VM status at all — during this work, questions as basic as "what is `vmRolloutStrategy` set to?" had to be answered by reading the module source in git rather than the live cluster.

That is a real ergonomic cost and it will recur constantly. The question is whether closing it is safe, and if so, for exactly which kinds.

## The constraint that makes this non-trivial

The role's safety rests on a **hand-curated apiGroup list** — described in its own review (#771) as *"a discipline guarantee, not a technical one."* Kinds were deliberately excluded (`helm.cattle.io`, `hub.traefik.io`, `monitoring.grafana.com`, `metallb.io/bgppeers`, specific external-secrets generator kinds) precisely because they can carry **inline plaintext credentials**, making a `get` equivalent to a Secret read. `kubevirt.io` has exactly that property, which is why it was left out rather than merely forgotten.

**Specifically: `VirtualMachine.spec.template.spec.volumes[].cloudInitNoCloud.userData` is an inline field.** If any VM used it, `get virtualmachine` would be a plain read of that VM's cloud-init — which for the Talos sandbox contains the **cluster CA private key**.

Both current VMs deliberately use `secretRef` instead, and both modules document that this is a convention rather than an incidental choice, citing this exact scenario. So the inline risk is currently unrealised — but it is enforced by discipline, not by the cluster. This is the same distinction the original review drew between `Middleware.spec.basicAuth` (secret-ref, accepted) and `Middleware.spec.headers.customRequestHeaders` (inline, recorded as residual risk).

## The bigger risk, which is not about inline secrets at all

**`subresources.kubevirt.io` must be excluded, and it is not obvious from the group name.** That group carries `virtualmachineinstances/console`, `/vnc`, `/portforward`, and the `start`/`stop`/`restart` subresources. A `get` on the console or VNC subresource is **interactive access to the guest**, not a read of metadata — the direct analogue of `pods/exec`, which the audit confirmed is absent everywhere in this role and must stay absent.

So any grant must be scoped to the `kubevirt.io` API group proper and must **not** be expressed in a way that sweeps in `subresources.kubevirt.io`. Watch for wildcard forms that would do this accidentally.

## Scope the research to per-kind decisions, not a blanket group grant

Enumerate what is actually installed and decide kind by kind. At least:

- `kubevirt.io` — `VirtualMachine` (inline userData risk above), `VirtualMachineInstance` (check what guest information lands in `status`), `VirtualMachineInstanceMigration`, `VirtualMachineInstancePreset`, `VirtualMachineInstanceReplicaSet`, and the `KubeVirt` CR itself (configuration; looks benign, confirm).
- `instancetype.kubevirt.io`, `pool.kubevirt.io`, `snapshot.kubevirt.io`, `clone.kubevirt.io`, `migrations.kubevirt.io` — check which exist on this cluster at all; don't grant for CRDs that aren't installed.
- `export.kubevirt.io` — **look at this one hard.** `VirtualMachineExport` mints tokens/URLs for exporting VM disk images. Even a read could expose an export token, which would be a disk-exfiltration path rather than an information leak. Default to excluding it.
- `cdi.kubevirt.io` — CDI is deliberately not installed here; confirm and skip.

For each: does any field, in `spec` **or** `status`, carry credentials, tokens, or anything that grants access rather than describing state?

## Options worth weighing

1. **Grant nothing** — status quo. Agents keep reading git. Cheapest, and the cost is only ergonomic.
2. **Grant a narrow enumerated set** (e.g. `VirtualMachineInstance` + `KubeVirt`, excluding `VirtualMachine`) — gets most of the operational visibility with none of the inline-userData exposure. Probably the sweet spot; verify that VMI status genuinely lacks the cloud-init content.
3. **Grant `VirtualMachine` too, and enforce no-inline-userData technically** rather than by discipline — a policy asserting that no VM sets `cloudInitNoCloud.userData`. This converts the discipline guarantee into a real one and is arguably worth doing *regardless* of the RBAC decision.
4. **Blanket `kubevirt.io: ["*"]` with read verbs** — matches the existing single wildcard rule's shape. Note that rule is verb- and apiGroup-limited and was found safe, but it is also the pattern this repo has been steadily moving away from.

## Verification standard

Per the standing rule that a control which looks correct but was never tested in the failing direction proves nothing: whatever is chosen must be verified by **injecting a VirtualMachine with inline `cloudInitNoCloud.userData` and confirming the MCP identity cannot read it** — not by observing that the role text looks right.

Two specific traps to avoid, both previously hit here:

- **If a Kyverno policy is used as the guard, do not write it in JMESPath for value inspection.** JMESPath operators glob-match when either operand contains `*`, which is how a `verbs: ["*"]` rule silently passed a read-only check once already. Use CEL.
- The app-level second layer (`read_only = true` and an explicit Secret deny in the MCP's own `config.toml`) is a genuine defence in depth, but it is not a substitute for correct RBAC and shouldn't be leaned on as one.

## Related

- #771 — the RBAC review that established the exclusion reasoning
- #816 — the `external-secrets.io` wildcard / fake-provider SecretStore question, the same class of problem in a different envelope
- ppat/homelab-ops-kubernetes-apps#3414 — original MCP RBAC hardening
- ppat/homelab-ops-kubernetes-apps#3503 — the KubeVirt/sandbox work that surfaced this, and whose Phase 6 this is now scoped under

## Done when

There is a decision per kind with reasoning recorded **in the role definition itself**, so the next person to touch it reads the file rather than this issue — including what would make the decision wrong later. If the answer is "grant nothing", that should be written down too, because otherwise this question simply gets re-asked the next time an agent can't see a VM.

ppat:
## Decision

`kubevirt.io` read **can** be added safely, but only for kinds that cannot reference a volume — which is a much
smaller set than the ticket assumed, and the ticket's preferred option 2 turns out not to work.

**Granted** (`get`/`list`/`watch`), enumerated by kind rather than folded into the group-wide `"*"` rule:

| Group | Kinds | Why safe |
| --- | --- | --- |
| `kubevirt.io` | `kubevirts` | Operator config + component health. No workload spec. This is the kind that answers the question in the ticket (`vmRolloutStrategy`). |
| `kubevirt.io` | `virtualmachineinstancemigrations` | Names a VMI plus a node selector; carries no VMI spec. |
| `kubevirt.io` | `virtualmachineinstancepresets` | `spec.domain` is a `DomainSpec`, which has no `volumes` field — the API server rejects `spec.domain.volumes` outright. |
| `instancetype.kubevirt.io` | all four | Structurally cannot carry cloud-init; a preference's `spec.volumes` is a single string, `preferredStorageClassName`. |
| `migrations.kubevirt.io` | `migrationpolicies` | Tuning knobs and a label selector. |

**Excluded**, with per-kind reasoning recorded in the role file itself rather than here. The deciding field is the
inline `VirtualMachine.spec.template.spec.volumes[].cloudInitNoCloud.userData`:

- `virtualmachines` — holds it directly.
- `virtualmachineinstances` — **option 2 does not work.** virt-controller copies the VM template into `VMI.spec`
  verbatim (`vmi.Spec = *vm.Spec.Template.Spec.DeepCopy()`), inline `userData` included. Granting the
  controller-written VMI instead of the human-written VM is the same plaintext in a different envelope, exactly
  like `ClusterGenerator` vs `Fake`. A `*/status` grant is no help either: `GET <kind>/status` returns the whole
  object.
- `virtualmachineinstancereplicasets`, `pool.kubevirt.io`, `snapshot.kubevirt.io`, `template.kubevirt.io`,
  `clone.kubevirt.io` — wrappers holding a VM spec, or free-form JSON patches against one.
  `VirtualMachineTemplate.spec.virtualMachine` is `x-kubernetes-preserve-unknown-fields` and stores a whole VM.
- `subresources.kubevirt.io` — serves `virtualmachineinstances/console`, `/vnc`, `/portforward` at verb **`get`**
  (KubeVirt derives the RBAC verb from the HTTP method, and those routes are GET), plus
  `virtualmachines/expand-spec` at `get`, which returns the fully expanded VM object and would hand back exactly
  the cloud-init the `virtualmachines` exclusion withholds.
- `subresources.template.kubevirt.io` — a second aggregated subresource group, new in v1.9.0.
- `export.kubevirt.io` — excluded, but not for the reason expected: `spec.tokenSecretRef` and
  `status.tokenSecretRef` are Secret *names*, and the export URLs are token-gated, so a read alone does **not**
  exfiltrate a disk. Excluded because nothing here creates exports, so it costs nothing, and "the token happens
  to be a reference in this version" is an implementation detail rather than a boundary.
- `backup.kubevirt.io`, `plugin.kubevirt.io` — new in v1.9.0, unused, unscrutinised status surface.
- `cdi.kubevirt.io` — not installed.

## The bug this uncovered

While proving the exclusion, the marker leaked anyway — through `apps/controllerrevisions`, which the role
already granted.

virt-controller writes a `ControllerRevision` holding the **entire `vm.spec`** into the VM's namespace on every
VM start. So a VM using inline `userData` puts that plaintext in apiGroup `apps`, with no `kubevirt.io` grant
involved at all. This is live on `homelab` today: the `mcp-kubernetes-readonly` identity can currently read
`talos-vm`'s full spec that way. Nothing sensitive comes back only because both VMs use `secretRef` — the
convention is doing all the work, and RBAC was doing none of it.

`apps/controllerrevisions` is removed from both roles. Cost: no read on StatefulSet/DaemonSet rollout history;
no MCP tool uses it.

## CI guard

Five new CEL rules in `ci/policy-tests/rbac-clusterrole-readonly.yaml` (CEL, not JMESPath — literal `"*"` must
compare as data, not glob):

- `no-apigroup-wildcard` — one `apiGroups: ["*"]` rule would silently re-admit every other exclusion in the file.
- `kubevirt-must-enumerate-kinds` — no resource wildcard on any `*.kubevirt.io` group. Uses `endsWith`, so it
  covers KubeVirt groups that do not exist yet.
- `kubevirt-excluded-groups` / `kubevirt-excluded-kinds` — the console/export groups and the VM-spec-bearing kinds.
- `no-controllerrevisions`.

Mutation-tested: eight fixtures, each violating exactly one rule, each failing exactly that rule; one clean
fixture passing all nine. A rule that only ever passes proves nothing.

## Verification

Full KubeVirt v1.9.0 installed on a disposable single-node cluster; a `VirtualMachine` created with inline
`cloudInitNoCloud.userData` carrying a marker string; the candidate ClusterRole applied verbatim and bound to a
probe ServiceAccount; every read performed as that identity.

Before the `controllerrevisions` fix, the marker came back in full through `apps/controllerrevisions`. After it,
a sweep across **every namespaced kind the identity can list** returns **0** occurrences of the marker, against
**13** for an unrestricted identity on the same objects — so the sweep is not vacuous. `console`, `vnc`,
`portforward` and `expand-spec` are refused both by `auth can-i` and by a real request. The KubeVirt CR
(`vmRolloutStrategy: Stage`), instancetypes and migrationpolicies stay readable, as do the virt-launcher Pod and
the VM's Events — which is what remains for answering "is this VM running, and where".

## Notes for review

- `nas` is unchanged apart from the `controllerrevisions` removal and a header update declaring the KubeVirt
  rules as a second permitted divergence. nas runs no KubeVirt and has no `*.kubevirt.io` CRD, so the
  strict-subset invariant holds.
- Two things worth knowing, neither addressed here: KubeVirt's own `kubevirt.io:default` ClusterRole is bound to
  `system:authenticated`, so `kubevirts` read was already available to every identity on the cluster; and
  `kubevirt.io:view` carries `aggregate-to-view`, so the `oidc-readers` binding already grants `homelab-users`
  read on `virtualmachines` and `virtualmachines/expand-spec`. That exposure is unchanged by this PR.

Closes #824

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -266,7 +266,18 @@ question: operators ship their own aggregating `ClusterRole`s (KubeVirt's
 `kubevirt.io:view` carries `rbac.authorization.k8s.io/aggregate-to-view`, and
 its `kubevirt.io:default` is bound to `system:authenticated`), so check what
 a subject already holds through `view` or through the operator's own bindings
-before concluding a kind is unreadable.
+before concluding a kind is unreadable. That also means an exclusion is a
+property of a *subject*, deliberately, not a cluster-wide guarantee: stock
+`system:aggregate-to-view` — which `view` consumes, and which `oidc-readers`
+binds `homelab-users` to — carries `apps/controllerrevisions` and `pods/log`
+alongside everything else it aggregates, so both channels a KubeVirt
+exclusion is built to withhold stay fully readable by any authenticated human
+through `view`. That's intentional: a dedicated read-only `ClusterRole` like
+this one gets scoped hardest around the identity reachable from an LLM — an
+agent that can be prompted, acts without a human reading every response, and
+can carry output past the cluster boundary — while an authenticated human
+holding `view` is a different threat model and is deliberately left
+unconstrained.
 
 Identities bound here must not live in `kube-system`: it's conventionally
 exempted from Pod Security Admission and from policy-engine namespace
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -250,11 +250,23 @@ another kind's content defeats a kind-level exclusion — wrapper kinds, or
 controller-written state/snapshot kinds that carry another kind's payload.
 When a group mixes credential-bearing and benign kinds, enumerate kinds
 rather than wildcarding the group, and re-check the enumeration whenever the
-operator adds a kind. `get`/`list`/`watch` is also not uniformly a read-only
-boundary: the API server maps HTTP method to RBAC verb for `*/proxy`
+operator adds a kind. Wrapper kinds do not respect API-group boundaries, so
+enumerate the wrappers before trusting an exclusion: KubeVirt records each
+`VirtualMachine`'s full spec in an `apps/controllerrevisions` object on every
+start, which puts the payload of an excluded kind in a group that looks
+entirely unrelated to it. `get`/`list`/`watch` is also not uniformly a
+read-only boundary: the API server maps HTTP method to RBAC verb for `*/proxy`
 subresources, and kubelet exposes GET routes for exec/attach/portForward, so
 `get` on `nodes/proxy` is code-execution-equivalent — treat proxy
-subresources as write access.
+subresources as write access. Aggregated APIs decide the verb the same way,
+which is why `subresources.kubevirt.io` serves a guest console at `get`.
+
+Finally, an exclusion is only meaningful for identities bound to the role in
+question: operators ship their own aggregating `ClusterRole`s (KubeVirt's
+`kubevirt.io:view` carries `rbac.authorization.k8s.io/aggregate-to-view`, and
+its `kubevirt.io:default` is bound to `system:authenticated`), so check what
+a subject already holds through `view` or through the operator's own bindings
+before concluding a kind is unreadable.
 
 Identities bound here must not live in `kube-system`: it's conventionally
 exempted from Pod Security Admission and from policy-engine namespace
--- a/ci/policy-tests/rbac-clusterrole-readonly.yaml
+++ b/ci/policy-tests/rbac-clusterrole-readonly.yaml
@@ -12,8 +12,11 @@
 # mcp-kubernetes-readonly.yaml on both clusters) before merge.
 #
 # Constrains what is INSIDE the mcp-kubernetes-readonly ClusterRole: no verb
-# outside get/list/watch, no nodes/proxy, no core Secrets, and an allow-list
-# on generators.external-secrets.io. The companion policy
+# outside get/list/watch, no nodes/proxy, no core Secrets, an allow-list on
+# generators.external-secrets.io, no "*" apiGroup, and — for KubeVirt — no
+# resource wildcard on any *.kubevirt.io group, no console/export subresource
+# groups, no kind that can embed a VM spec, and no apps/controllerrevisions
+# (which is one such kind in disguise). The companion policy
 # rbac-clusterrolebinding-roleref.yaml (test-rbac-clusterrolebinding-roleref)
 # constrains which role is actually BOUND to the mcp-kubernetes ServiceAccount
 # — this file says nothing about that. Both must pass for the read-only
@@ -147,3 +150,136 @@ spec:
             denying fakes/webhooks alone does not deny either wrapper. Use
             an explicit resource allow-list instead of "*", and keep
             fakes/webhooks/clustergenerators/generatorstates off it.
+  - name: no-apigroup-wildcard
+    match:
+      any:
+      - resources:
+          kinds:
+          - ClusterRole
+    validate:
+      cel:
+        expressions:
+        - expression: "!object.rules.exists(r, '*' in r.apiGroups)"
+          message: >-
+            A rule uses "*" in apiGroups. Every other rule in this policy
+            excludes things by naming an apiGroup — core Secrets, the
+            generators.external-secrets.io allow-list, the KubeVirt rules
+            below — so a single apiGroups: ["*"] rule silently re-admits all
+            of them at once and makes the rest of this policy decorative. It
+            also re-admits every group nobody has reviewed yet, including
+            ones installed after this file was last read. The role's own
+            group-wide grant already demonstrates the correct shape: a
+            wildcard on RESOURCES within an explicitly enumerated list of
+            apiGroups. Enumerate the apiGroups.
+  - name: kubevirt-must-enumerate-kinds
+    match:
+      any:
+      - resources:
+          kinds:
+          - ClusterRole
+    validate:
+      cel:
+        expressions:
+        - expression: >-
+            !object.rules.exists(r,
+            r.apiGroups.exists(g, g == 'kubevirt.io' || g.endsWith('.kubevirt.io')) &&
+            ('*' in r.resources))
+          message: >-
+            A rule wildcards resources on a KubeVirt apiGroup. KubeVirt
+            spreads across 11 groups that mix benign configuration
+            (KubeVirt CR, instancetypes, migrationpolicies) with kinds that
+            embed a full VM spec — and therefore any inline
+            cloudInitNoCloud.userData — plus, in subresources.kubevirt.io, a
+            guest console. Its API surface also grows across minors: v1.9.0
+            added backup.kubevirt.io, template.kubevirt.io,
+            plugin.kubevirt.io and a second aggregated subresource group, so
+            a "*" written today grants kinds nobody has read the schema of.
+            The endsWith() match is deliberate: it covers KubeVirt groups
+            that do not exist yet. Enumerate the resources by kind.
+  - name: kubevirt-excluded-groups
+    match:
+      any:
+      - resources:
+          kinds:
+          - ClusterRole
+    validate:
+      cel:
+        expressions:
+        - expression: >-
+            !object.rules.exists(r,
+            r.apiGroups.exists(g, g in ['subresources.kubevirt.io',
+            'subresources.template.kubevirt.io', 'export.kubevirt.io',
+            'cdi.kubevirt.io']))
+          message: >-
+            A rule names a KubeVirt apiGroup that must never appear in a
+            read-only role. subresources.kubevirt.io serves
+            virtualmachineinstances/console, /vnc and /portforward, and
+            KubeVirt derives the RBAC verb from the HTTP method — those
+            routes are GET, so "get" there is an interactive session inside
+            the guest, exactly like the nodes/proxy case above. The same
+            group also serves virtualmachines/expand-spec at verb "get",
+            which returns the fully expanded VM object and would hand back
+            the inline cloud-init that excluding virtualmachines exists to
+            withhold. subresources.template.kubevirt.io is a second
+            aggregated subresource group (new in v1.9.0).
+            export.kubevirt.io is the VM disk-image export control plane.
+            cdi.kubevirt.io is not installed. Remove the group.
+  - name: kubevirt-excluded-kinds
+    match:
+      any:
+      - resources:
+          kinds:
+          - ClusterRole
+    validate:
+      cel:
+        expressions:
+        - expression: >-
+            !object.rules.exists(r,
+            r.apiGroups.exists(g, g == 'kubevirt.io' || g.endsWith('.kubevirt.io')) &&
+            r.resources.exists(res, res in ['virtualmachines',
+            'virtualmachineinstances', 'virtualmachineinstancereplicasets',
+            'virtualmachinesnapshotcontents', 'virtualmachinetemplates',
+            'virtualmachinepools', 'virtualmachineclones',
+            'virtualmachinerestores', 'virtualmachineexports']))
+          message: >-
+            A rule grants a KubeVirt kind that can carry inline plaintext
+            cloud-init. VirtualMachine holds
+            spec.template.spec.volumes[].cloudInitNoCloud.userData directly;
+            the rest are wrappers for it, and RBAC allows and denies by
+            kind, never by content. VirtualMachineInstance is the rendered
+            copy of the VM template, userData included, so granting it
+            instead of VirtualMachine is not a mitigation.
+            VirtualMachineInstanceReplicaSet and VirtualMachinePool hold VM
+            templates; VirtualMachineSnapshotContent embeds a whole
+            VirtualMachine; VirtualMachineTemplate.spec.virtualMachine is
+            x-kubernetes-preserve-unknown-fields and stores one verbatim;
+            VirtualMachineClone and VirtualMachineRestore carry free-form
+            JSON patches that can set any field on the target.
+            VirtualMachineExport is the disk-export control plane. Grant the
+            KubeVirt CR, instancetypes/preferences, migrationpolicies,
+            virtualmachineinstancemigrations and
+            virtualmachineinstancepresets instead — none of those can
+            reference a volume at all.
+  - name: no-controllerrevisions
+    match:
+      any:
+      - resources:
+          kinds:
+          - ClusterRole
+    validate:
+      cel:
+        expressions:
+        - expression: "!object.rules.exists(r, ('apps' in r.apiGroups) && ('controllerrevisions' in r.resources))"
+          message: >-
+            A rule grants apps/controllerrevisions. That is a wrapper kind
+            for a workload spec, and on a cluster running KubeVirt it is a
+            complete bypass of the kubevirt-excluded-kinds rule above:
+            virt-controller writes a ControllerRevision whose .data is the
+            entire vm.spec every time a VirtualMachine starts, so a VM using
+            inline cloudInitNoCloud.userData puts that plaintext in apiGroup
+            "apps" with no kubevirt.io grant involved. Verified by reading
+            one on the homelab cluster through the mcp-kubernetes-readonly
+            identity itself. Excluding the KubeVirt kinds while granting
+            this achieves nothing. Remove controllerrevisions from
+            resources; nothing but StatefulSet/DaemonSet rollout history
+            reads them.
--- a/ci/policy-tests/rbac-clusterrole-readonly.yaml
+++ b/ci/policy-tests/rbac-clusterrole-readonly.yaml
@@ -209,7 +209,9 @@ spec:
             !object.rules.exists(r,
             r.apiGroups.exists(g, g in ['subresources.kubevirt.io',
             'subresources.template.kubevirt.io', 'export.kubevirt.io',
-            'cdi.kubevirt.io']))
+            'cdi.kubevirt.io', 'snapshot.kubevirt.io', 'template.kubevirt.io',
+            'pool.kubevirt.io', 'clone.kubevirt.io', 'backup.kubevirt.io',
+            'plugin.kubevirt.io']))
           message: >-
             A rule names a KubeVirt apiGroup that must never appear in a
             read-only role. subresources.kubevirt.io serves
@@ -223,7 +225,16 @@ spec:
             withhold. subresources.template.kubevirt.io is a second
             aggregated subresource group (new in v1.9.0).
             export.kubevirt.io is the VM disk-image export control plane.
-            cdi.kubevirt.io is not installed. Remove the group.
+            cdi.kubevirt.io is not installed. snapshot.kubevirt.io,
+            template.kubevirt.io, pool.kubevirt.io and clone.kubevirt.io each
+            embed a whole VirtualMachine or a free-form JSON patch that can
+            set any field on one, same as the kinds kubevirt-excluded-kinds
+            rejects above — see the role file's own "KubeVirt API groups
+            deliberately granted NOTHING" comment for the per-group
+            reasoning, including why virtualmachinetemplaterequests
+            (installed on homelab today) is excluded by group rather than by
+            kind. backup.kubevirt.io and plugin.kubevirt.io carry no VM spec
+            but are excluded as unused and unaudited. Remove the group.
   - name: kubevirt-excluded-kinds
     match:
       any:
@@ -236,14 +247,17 @@ spec:
         - expression: >-
             !object.rules.exists(r,
             r.apiGroups.exists(g, g == 'kubevirt.io' || g.endsWith('.kubevirt.io')) &&
-            r.resources.exists(res, res in ['virtualmachines',
+            r.resources.exists(res, ['virtualmachines',
             'virtualmachineinstances', 'virtualmachineinstancereplicasets',
             'virtualmachinesnapshotcontents', 'virtualmachinetemplates',
             'virtualmachinepools', 'virtualmachineclones',
-            'virtualmachinerestores', 'virtualmachineexports']))
+            'virtualmachinerestores', 'virtualmachineexports'].exists(k,
+            res == k || res.startsWith(k + '/'))))
           message: >-
             A rule grants a KubeVirt kind that can carry inline plaintext
-            cloud-init. VirtualMachine holds
+            cloud-init, either directly or via its */status subresource (a
+            GET on <kind>/status returns the whole object, spec included, so
+            it is not a narrower grant). VirtualMachine holds
             spec.template.spec.volumes[].cloudInitNoCloud.userData directly;
             the rest are wrappers for it, and RBAC allows and denies by
             kind, never by content. VirtualMachineInstance is the rendered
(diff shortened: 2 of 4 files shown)


ppat:
## Security review — `9dd57cf`

> **Updated after review discussion.** Disposition agreed and recorded below: findings 2, 3 and 4 are fixed in this PR; finding 1 is deferred to **#843** with the risk explicitly accepted. Findings and evidence are unchanged from the original comment.

The core reasoning holds up. I re-derived the load-bearing claims independently rather than accepting them: the `controllerrevisions` bug is real and I reproduced it on `homelab`, the structural "cannot carry a volume" arguments for the granted kinds check out against the v1.9.0 Go types, and the nas ⊆ homelab invariant holds programmatically. The wrapper-kind framing is the right frame, and finding the `apps/controllerrevisions` leak while proving an exclusion is the kind of result that only comes from actually testing the boundary.

Four findings. The first is the significant one: **the guest-console exclusion, which the PR treats as one of its two hard boundaries, does not hold end-to-end** — the same data is already reachable through a core kind this role has granted since #771. The other three are about the CI guard being narrower than the exclusions it is presented as enforcing, verified by mutation-testing it the same way you did.

**Method.** Kyverno CLI v1.18.2 (the pinned version) run locally against the PR's own files and against hand-built mutation fixtures; KubeVirt v1.9.0 CRD/Go API types pulled from upstream at that exact tag; live reads against `homelab` for everything claimed about the running cluster. Both role files pass 9/9 under the new policy locally, matching green CI.

---

## Expected in this PR before it lands

- [ ] **`kubevirt-excluded-kinds` — apply the `startsWith` patch** (finding 2). Verified: catches the `virtualmachines/status` fixture (`pass: 8, fail: 1`) and leaves both real role files at 9/9.
- [ ] **`kubevirt-excluded-groups` — add the six remaining groups** `snapshot` / `template` / `pool` / `clone` / `backup` / `plugin` `.kubevirt.io` (finding 3), so the guard matches the exclusion set the role file's prose already claims. None of them is granted anything, so this is closing the gap rather than changing a grant.
- [ ] **Fixtures for both new behaviours**, each failing exactly one rule, to the standard the PR already sets. The two that matter: the `<kind>/status` form, and a rule naming an uncovered group (`template.kubevirt.io/virtualmachinetemplaterequests` is the sharp one — that CRD is installed on `homelab` today).
- [ ] **DESIGN.md — one sentence** (finding 4): stock `system:aggregate-to-view` also carries `apps/controllerrevisions` and `pods/log`, so an exclusion is a property of a *subject*; `view` is the other subject on this cluster.
- [ ] **Comment amendment in both role files, documentation only** — the `subresources.kubevirt.io` exclusion should say what it actually buys: no interactive session and no arbitrary VM-spec read, but *not* console **output**, which reaches this identity via `pods/log`, tracked in #843. No RBAC change. This is the whole mitigation for finding 1 within this PR's scope, and it matters because this file's value rests on its comments being currently true rather than aspirational.
- [ ] **Housekeeping** — the PR body's "eight fixtures / nine rules" line will need updating, and the first commit reworded to fold in the later ones before the PR is updated (GitHub seeds the squash commit from it).

**Deferred to #843:** finding 1, the guest serial console via `pods/log`. Not created by this PR (`pods/log` has been granted since #771), not bound to the kinds this PR introduces, and the fix is a scoping question — cluster-wide `pods/log` → namespace `Role`s — that would swamp the PR. Interim risk accepted: both VMs are sandboxes holding no valuable data, work is in progress to destroy and recreate them weekly including their PVCs, and nothing currently printed to either console is a credential. #843 also records why the KubeVirt-side knob was rejected, so it doesn't get re-proposed.

---

### 1. The guest serial console is readable via `pods/log`, so the `subresources.kubevirt.io` exclusion is not the boundary it is described as — deferred to #843 (Medium–High)

The role file's stated reason for excluding `subresources.kubevirt.io` is that `virtualmachineinstances/console` at verb `get` "is an interactive session inside the guest, the direct analogue of pods/exec". That reasoning is correct. But KubeVirt v1.9 streams the same serial console to a **container log** by default, and `pods/log` is granted cluster-wide by this role.

Every virt-launcher pod on `homelab` today runs a `guest-console-log` sidecar (`virt-tail --logfile .../virt-serial0-log`), because `logSerialConsole` defaults to true. Reading that container's log returns the guest's console stream. Verified live, `sandbox-docker`:

```
$ kubectl logs virt-launcher-docker-vm-7s69q -c guest-console-log
[   36.021279] cloud-init[1337]: mise: installing mise...
[   38.610582] cloud-init[1337]: mise cargo-binstall@1.21.1 [2/3] verify GitHub artifact attestations
ci-info: ++++++++++Authorized keys from /home/docker/.ssh/authorized_keys for user docker++++++++++
ci-info: | ssh-ed25519 | de:4c:67:f7:00:5b:...:47:e3:0a:7b | - | sandbox-docker-vm |
cloud-init: -----BEGIN SSH HOST KEY FINGERPRINTS-----
cloud-init: 256 SHA256:uCqIXbsfvNMxdq8RPjy/CAiUFNFezWbNt+I8JtcGxt4 root@docker-vm (ED25519)
-----BEGIN SSH HOST KEY KEYS-----
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIFu5PJ... root@docker-vm
Cloud-init v. 26.1-0ubuntu2 finished at Sun, 02 Aug 2026 20:31:36 +0000. Datasource DataSourceNoCloud [seed=/dev/vdc].
```

`sandbox-talos` streams the same way (kernel ring buffer, CNI churn, `[talos] bootstrap request received`). The `compute` container's own log is clean at the current verbosity — no domain XML — so this is specifically the console sidecar.

**Why this matters for this PR specifically.** The whole exclusion set is built to keep this identity out of the guest and away from anything the VM's cloud-init produces. What is readable today is not catastrophic — host public keys are public by construction, and fingerprints of authorized keys are not credentials. But the channel is continuous in-guest output, and its content is whatever the guest prints. `runcmd`/`bootcmd` output is echoed verbatim, as the `mise` install lines show. A VM whose cloud-init curls an artifact with a token in the URL, or echoes anything during provisioning, puts it here — in `pods/log`, in a core apiGroup, with no `kubevirt.io` grant involved. That is structurally identical to the `apps/controllerrevisions` finding this PR is built around, and it is live rather than latent.

**Why the verification missed it.** The marker sweep was over objects the identity can *list*, and pod logs are not an object. It would also have come back clean even if logs had been swept: cloud-init does not echo `userData` back to the console, it echoes the *results* of running it. The marker test and this channel are orthogonal — a clean sweep is not evidence about it either way.

**Disposition — deferred to #843, doc fix here.** The KubeVirt-side knobs (`disableSerialConsoleLog` cluster-wide, `logSerialConsole: false` per VM) are rejected: they remove the console log for **human cluster admins too**, and `virtctl console` is not a substitute because it attaches live and cannot show the boot log of a VM that already failed. That leaves who-can-read-`pods/log` as the only lever, which is an RBAC scoping project rather than a change to this PR — and one that revisits a decision #771 recorded deliberately. #843 carries the full analysis, including why the VM namespaces being dedicated and a Kyverno `generate` inversion may make it cheaper than #771 assumed. What lands here is the comment amendment above, so the role file stops implying a boundary it does not hold.

### 2. `kubevirt-excluded-kinds` is defeated by the `/status` form — the exact bypass the role file documents (Medium)

The role file already states the mechanism: *"a `*/status` subresource grant is no help either: a `GET <kind>/status` returns the whole object, spec included."* The CEL rule does not implement it. `res in ['virtualmachines', ...]` is exact string equality, and `virtualmachines/status` is a different string.

Mutation-tested against the PR's policy as merged:

```yaml
- apiGroups: [kubevirt.io]
  resources: [kubevirts, virtualmachines/status, virtualmachineinstances/status]
  verbs: [get, list, watch]
```

→ `pass: 9, fail: 0`. The control fixture (bare `virtualmachines`) correctly fails `kubevirt-excluded-kinds`. So the guard catches the naive edit and misses the one that returns the same bytes.

This isn't a contrived shape either — this role grants `pods/status`, `deployments/status`, `daemonsets/status`, `nodes/status` and a dozen more, so `<kind>/status` is the file's own idiom. Someone adding "just the status, not the spec" is the most likely regression here, and it's the one that passes.

Verified fix — catches the fixture (`pass: 8, fail: 1`), leaves both real files at 9/9:

```cel
!object.rules.exists(r,
r.apiGroups.exists(g, g == 'kubevirt.io' || g.endsWith('.kubevirt.io')) &&
r.resources.exists(res, ['virtualmachines',
'virtualmachineinstances', 'virtualmachineinstancereplicasets',
'virtualmachinesnapshotcontents', 'virtualmachinetemplates',
'virtualmachinepools', 'virtualmachineclones',
'virtualmachinerestores', 'virtualmachineexports'].exists(k,
res == k || res.startsWith(k + '/'))))
```

I used `startsWith` rather than `split('/')[0]` deliberately — no dependency on which CEL extension libraries Kyverno has loaded. Worth the same treatment on `no-controllerrevisions` for consistency, though ControllerRevision has no status subresource so nothing turns on it there.

### 3. The CEL exclusion set is narrower than the exclusion set the role file claims (Low–Medium)

The role file lists eight KubeVirt groups as "deliberately granted NOTHING". `kubevirt-excluded-groups` enforces four of them (`subresources.kubevirt.io`, `subresources.template.kubevirt.io`, `export.kubevirt.io`, `cdi.kubevirt.io`), and `kubevirt-excluded-kinds` catches most of the rest by kind — but not all. This fixture passes 9/9:

```yaml
- apiGroups: [template.kubevirt.io]
  resources: [virtualmachinetemplaterequests]     # CRD exists on homelab today
- apiGroups: [backup.kubevirt.io]
  resources: [virtualmachinebackups, virtualmachinebackuptrackers]
- apiGroups: [snapshot.kubevirt.io]
  resources: [virtualmachinesnapshots]
- apiGroups: [pool.kubevirt.io]
  resources: [virtualmachinepools/status]         # also finding 2
```

`virtualmachinetemplaterequests.template.kubevirt.io` is installed on `homelab` right now and appears in neither CEL list; the file's argument for excluding `template.kubevirt.io` rests on `VirtualMachineTemplate` embedding a whole VM, and the request kind sits in the same group with no equivalent scrutiny recorded. (Its Go types aren't in `kubevirt/kubevirt` staging at v1.9.0 — it ships from the separate virt-template component — so I couldn't confirm its schema from source, which is itself an argument for excluding it by group rather than reasoning about the kind.)

Since none of these groups is granted anything at all, the cheap fix is to move them into `kubevirt-excluded-groups` wholesale — `snapshot`, `template`, `pool`, `clone`, `backup`, `plugin` — rather than maintaining a kind list that has to track each group's inventory across minors. That also makes the guard match the file's prose exactly, which right now it doesn't: the comment says CI "fails the build if any of the first three reappears", which is accurate but sets a much lower bar than a reader of the surrounding exclusion commentary would assume.

### 4. The closure is identity-scoped, and `view` still carries both channels (context)

The PR notes `kubevirt.io:view` carries `aggregate-to-view`. Confirmed live, and it's broader than the note suggests — `kubevirt.io:view` grants `virtualmachines`, `virtualmachineinstances`, `virtualmachineinstancereplicasets`, snapshot/backup/export/clone/pool kinds, and `virtualmachines/expand-spec` at `get`. Since `oidc-readers` binds `homelab-users` → `view`, that group holds today exactly what this PR withholds from the MCP identity.

Worth adding: stock `system:aggregate-to-view` also grants **`apps/controllerrevisions`** and **`pods/log`**. So both channels this review is about — the ControllerRevision holding the full `vm.spec`, and the guest console log — remain readable by every `homelab-users` reader. The title says "close the controllerrevisions bypass"; it's closed for one ServiceAccount. That's a defensible scope (the MCP identity is the one reachable from an LLM), but the DESIGN.md paragraph added here is precisely the place to say the quiet part: an exclusion is a property of a *subject*, and `view` is the other subject on this cluster.

Resolution: the one-sentence DESIGN.md addition above. Whether `homelab-users` should keep `view` on a cluster now running VMs is out of scope here — #843 records the same split, and #838 already records it for `kubevirt.io:view`.

---

### Verified and agreed

- **The `controllerrevisions` bug is real.** `revision-start-vm-16368915-...-1` in `sandbox-talos`, owner-referenced to `VirtualMachine/talos-vm`, `.data.spec` holding the entire VM spec — `runStrategy`, full `domain`, node affinity, and `volumes[].cloudInitNoCloud.secretRef`. Exactly as described, and today `secretRef` is doing all the work. Removing the grant is right, and finding it while proving an unrelated exclusion is the strongest part of this PR.
- **`kubevirt.io:default` is bound to `system:authenticated` and grants `get`/`list` on `kubevirts`.** Confirmed both objects. So the `kubevirts` grant adds only `watch` for this identity — no new exposure, and the note in the PR body is accurate.
- **The preset argument holds.** `DomainSpec` at v1.9.0 has no `Volumes` field (`Resources`, `CPU`, `Memory`, `Machine`, `Firmware`, `Clock`, `Features`, `Devices`, `IOThreads*`, `Chassis`, `LaunchSecurity`, `RebootPolicy`). Structurally cannot carry cloud-init, as claimed.
- **The instancetype argument holds.** `VirtualMachineInstancetypeSpec` carries no volume content, and `VolumePreferences` is exactly one field, `preferredStorageClassName`.
- **`migrationpolicies` holds.** Selectors plus tuning knobs, no workload spec.
- **nas ⊆ homelab holds.** Expanded both roles to `(apiGroup, resource, verb)` triples: homelab 391, nas 355, **zero** nas-only triples. Groups present on homelab and absent from nas are exactly the four declared plus the three KubeVirt ones — matching the header. That's ~15 lines of Python and it's the one invariant no per-file policy can express; happy to drop it in as a pre-commit hook if you want it (it's the check the #771 review asked for and is still the only unguarded property here).
- **CI reproduces.** Both files, 9/9, Kyverno v1.18.2 locally.

### Minor accuracy notes

- **`preferredLaunchSecurity`.** The SEV note names `spec.launchSecurity.sev.{dhCert,session}` on an instancetype. Preferences carry the same `*v1.LaunchSecurity` type as `spec.preferredLaunchSecurity`, and both preference kinds are granted. The accepted-risk argument (SEV session material is designed to transit an untrusted host; no SEV hardware here) applies unchanged — the field list is just incomplete.
- **`virtualmachineinstancemigrations` has grown.** The description "names a VMI plus a node selector" predates v1.9: the spec now also has `sendTo.connectURL` (decentralized-migration synchronization controller URL), `receive.migrationID` and `priority`, and the status has `synchronizationAddresses`. Still no VMI spec and nothing credential-bearing, so the grant stands — but "carries no VMI spec of its own" is the durable claim, not the field inventory.

ppat:
Findings 2, 3 and 4 addressed in dabf2fa, plus the two minor accuracy notes. Finding 1 is deliberately untouched here, per operator decision — deferred to #843, not missed.

## Finding 2 — `/status` bypass

Applied the `startsWith` fix to `kubevirt-excluded-kinds` exactly as you verified it, and extended the same treatment to `no-controllerrevisions` for consistency (agreed it's not load-bearing there — ControllerRevision has no status subresource — but the guard should look the same everywhere it exists for the same reason).

Measured, mutation-testing your fixture against the changed policy: `virtualmachines/status` + `virtualmachineinstances/status` goes from **pass 9/fail 0 → pass 8/fail 1**, matching your numbers exactly. Both real role files stay at **9/9** (measured, both clusters).

## Finding 3 — narrower-than-claimed group exclusion

Added the six groups to `kubevirt-excluded-groups`: `snapshot`, `template`, `pool`, `clone`, `backup`, `plugin` (`.kubevirt.io`). None of them was granted anything before, so this is closing the gap, not changing a grant, as you said.

Your four-rule fixture (`template.kubevirt.io/virtualmachinetemplaterequests`, `backup.kubevirt.io/{virtualmachinebackups,virtualmachinebackuptrackers}`, `snapshot.kubevirt.io/virtualmachinesnapshots`, `pool.kubevirt.io/virtualmachinepools/status`), measured: **pass 9/fail 0 → pass 7/fail 2**. Two rules trip it, not one — `kubevirt-excluded-groups` on all four lines, and `kubevirt-excluded-kinds` on the `virtualmachinepools/status` line, which you'd already flagged as double-covered by finding 2's fix. Expected, not a new problem.

Fixed the role file's stale claim that CI "fails the build if any of the first three reappears" — it now names the actual three rules and what each one covers post-fix. Also added the `virtualmachinetemplaterequests` case inline on the `template.kubevirt.io` bullet, since that's the concrete argument for excluding by group.

One correction on your count: the role file's "deliberately granted NOTHING" footer lists **ten** groups, not eight — I count `subresources.kubevirt.io`, `subresources.template.kubevirt.io`, `export.kubevirt.io`, `cdi.kubevirt.io`, plus the six you named. The action item was unambiguous either way (enforce the six missing ones), so it didn't change what got fixed, but flagging it since the whole point of this exercise is comment accuracy.

## Finding 4 — exclusion is a property of a subject

Landed as one sentence, but not the one you drafted, on operator instruction: `oidc-readers`/`homelab-users`/`view` is explicitly out of scope for narrowing, now and later — humans reading `view` are a deliberately different, unconstrained threat model, not a gap sharing the same fix path as the MCP identity. So the DESIGN.md addition doesn't just note the asymmetry, it says why it's intentional: stock `system:aggregate-to-view` also carries `apps/controllerrevisions` and `pods/log`, both channels stay fully readable by any human through `view`, and that's correct, because `mcp-kubernetes-readonly` is scoped hardest around the identity reachable from an LLM specifically. No RBAC change to any human-facing subject, and no follow-up issue for that path — there's nothing to follow up.

## Comment amendment (doc-only, finding 1's in-scope piece)

Updated the `subresources.kubevirt.io` bullet in `clusters/homelab`'s role file: excluding it buys no interactive session and no arbitrary VM-spec read via `expand-spec`, but not console *output* — the `guest-console-log` sidecar streams the same serial console to a container log by default, and `pods/log` is granted cluster-wide above regardless of this exclusion. References #843, doesn't attempt its fix.

One thing worth flagging: your checklist said "in both role files." `clusters/nas`'s copy has no `subresources.kubevirt.io` commentary at all — nas runs no KubeVirt, so there was nothing there to amend. I only touched the homelab copy; if you meant something else by "both," let me know.

## Minor accuracy notes — both fixed

- `preferredLaunchSecurity`: the instancetype comment now names both the instancetype and preference fields (`*v1.LaunchSecurity` via `spec.preferredLaunchSecurity`), since both preference kinds are granted. Accepted-risk argument unchanged.
- `virtualmachineinstancemigrations`: rewritten so "carries no VMI spec of its own" is the claim the comment rests on, with the v1.9 field additions (`sendTo.connectURL`, `receive.migrationID`, `priority`, `status.synchronizationAddresses`) noted as illustrative rather than exhaustive, so it doesn't go stale again next minor.

## Verification

Re-ran the full `ci/policy-tests` suite the way CI actually invokes it (`kyverno apply` against each cluster's whole `cluster/rbac/` directory, both policy files together) after all changes: **12/12 pass on both clusters**, unchanged from before the fix — no regression from touching four CEL expressions and two comment blocks in the same pass.

## nas ⊆ homelab pre-commit hook

Not added — real gap, genuinely useful offer, but it's a new mechanism the operator hasn't asked for. Recording it here as an open offer rather than acting on it.

---

Housekeeping items from your checklist (PR body's "eight fixtures / nine rules" line, squashing the commit history) weren't touched — out of scope for this pass; the operator can fold that in at merge time.

Pushed to `feat/mcp-rbac-kubevirt` as `dabf2fa`.

PR #842, opened 2026-08-03
feat(cluster-homelab): add sandbox-lifecycle namespace, RBAC, and a Delete-reclaim StorageClass

ppat:
## Summary

Clusters-repo half of the weekly sandbox VM destroy-and-rebuild mechanism
(ppat/homelab-ops-kubernetes-experiments#227). The CronJobs themselves ship in
a companion PR against that repo
(`experiments/apps/sandbox-lifecycle/`) — this PR is the "cage": the
namespace, isolation, and RBAC boundary that lets those CronJobs reach this
host cluster's API from a namespace AI agents cannot write to, plus the
StorageClass fix that must exist before either sandbox VM's PVC can safely be
destroyed on a schedule.

- **`sandbox-lifecycle` namespace + NetworkPolicy** — default-deny both
  directions, egress limited to CoreDNS and this cluster's own kube-apiserver
  ClusterIP (`10.43.0.1/32:443`), no ingress rule at all. Split out from
  `sandbox-docker`/`sandbox-talos` because those two deliberately *cannot*
  reach the host API (their egress policy excepts it, and the live
  falsifiability probe asserts that stays blocked every cycle) — a
  host-API-capable ServiceAccount can only safely live somewhere agents
  cannot create pods.
- **`vm-rebuilder` ServiceAccount** in `sandbox-lifecycle`, plus one
  `Role`+`RoleBinding` each in `sandbox-docker`/`sandbox-talos`
  (`role-sandbox-docker.yaml`, `role-sandbox-talos.yaml`), every verb scoped
  to the one named VM/PVC except `pods:list` (unavoidable — the virt-launcher
  pod name is generated). `sandbox-talos`'s Role additionally grants `create`
  on `jobs`, needed to trigger `bootstrap-talos-vm`/`provision-agent-admin`
  after a destroy — see that file's own comment for the full threat-model
  reasoning on why this was judged an acceptable grant (it equals what AI
  agents already hold in that namespace; the SA holding it lives somewhere
  agents can't reach).
- **`sc-longhorn-local-non-replicated-ephemeral`** — identical to the existing
  `sc-longhorn-local-non-replicated` except `reclaimPolicy: Delete`. The
  existing class is `Retain`: deleting a PVC on it leaves the PV `Released`
  and the Longhorn volume intact, so a weekly destroy against either sandbox
  PVC as declared today would leak 200Gi/40Gi of storage every single week
  while every other check reports success. `storageClassName` is immutable on
  an existing PVC, so this class has to exist, and each PVC has to be
  migrated onto it via a real destroy/recreate (not a live patch), before
  either rebuild CronJob is safe to unsuspend — see the companion PR's
  module README for the exact migration sequence. This PR ships the class
  only; migrating the PVCs is a separate, deliberate step.
- **`config-services-sandbox-lifecycle` Kustomization** — 1m0s interval
  (matches `sandbox-docker`/`sandbox-talos`'s own fast-reconcile isolation
  Kustomizations), depending on `infra-security-core`, `infra-networking-core`,
  and both existing sandbox Kustomizations (the Roles here target namespaces
  those own, so those namespaces must exist first).

## Ordering / what happens if this merges before or after the CronJobs PR

This PR has zero blast radius on its own — it's pure addition (new namespace,
new RBAC, new StorageClass) and nothing references any of it yet. The
companion CronJobs PR ships both CronJobs with `spec.suspend: true` and each
script's own Step 0 refuses to run against a PVC still on the `Retain` class,
so even an out-of-order merge (CronJobs before this PR, or either PR merged
without the other) cannot silently leak storage — the failure mode, if any,
is a loud, immediate "refusing to destroy" or an RBAC-denied error, never a
silent leak.

## Test plan

- [x] `pre-commit run --all-files` clean (yamllint, markdownlint, shellcheck,
      kubeconform via `validate-kubernetes-manifests`)
- [x] `kubectl kustomize clusters/homelab/services/sandbox-lifecycle` and
      `kubectl kustomize clusters/homelab` (full tree) both build cleanly;
      confirmed the Roles land in `sandbox-docker`/`sandbox-talos` and the
      RoleBindings' subjects correctly cross-reference the SA in
      `sandbox-lifecycle`
- [ ] Live reconcile (out of scope for this session — no destroy was
      performed against either sandbox VM; see companion PR for what remains
      unproven)

Part of ppat/homelab-ops-kubernetes-experiments#227 (that repo's companion PR
carries the closing keyword; this repo has no local issue for #227).

### DIFF
--- a/clusters/homelab/kustomizations/config-services-sandbox-lifecycle.yaml
+++ b/clusters/homelab/kustomizations/config-services-sandbox-lifecycle.yaml
@@ -1,0 +1,33 @@
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: config-services-sandbox-lifecycle
+  namespace: flux-system
+spec:
+  dependsOn:
+  - name: infra-security-core
+    namespace: flux-system
+  - name: infra-networking-core
+    namespace: flux-system
+  # The Roles this module ships live in sandbox-docker/sandbox-talos (see
+  # role-sandbox-docker.yaml, role-sandbox-talos.yaml -- a Role is only valid once its
+  # namespace exists), so those two namespaces' own Kustomizations are hard prerequisites,
+  # not just something that happens to converge first.
+  - name: config-services-sandbox-docker
+    namespace: flux-system
+  - name: config-services-sandbox-talos
+    namespace: flux-system
+  # Short interval, split out of config-services.yaml's 15m0s on purpose, same reasoning as
+  # sandbox-docker/sandbox-talos's own Kustomizations: this module IS an isolation/RBAC
+  # boundary (the ServiceAccount and Roles that let the weekly rebuild CronJobs reach this
+  # host cluster's API), so drift correction needs to converge fast.
+  interval: 1m0s
+  path: ./clusters/homelab/services/sandbox-lifecycle
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: root
+    namespace: flux-system
+  timeout: 2m0s
+  wait: true
--- a/clusters/homelab/kustomizations/kustomization.yaml
+++ b/clusters/homelab/kustomizations/kustomization.yaml
@@ -26,5 +26,6 @@ resources:
 - config-services.yaml
 - config-services-sandbox-docker.yaml
 - config-services-sandbox-talos.yaml
+- config-services-sandbox-lifecycle.yaml
 - policy-pod-security-standards.yaml
 - policy-best-practices.yaml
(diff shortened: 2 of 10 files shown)


ppat:
Follow-up commit: clarified role-sandbox-talos.yaml's comment — the
`bootstrap-talos-vm` CronJob its `cronjobs` rule names is forward-declared,
not live yet. The companion experiments-repo PR
(ppat/homelab-ops-kubernetes-experiments#233) originally converted
`sandbox-talos/bootstrap-job.yaml` into that CronJob directly, but that
directory is owned by the in-flight agent-admin credential relay (#225), so
that conversion was pulled back out and is now just documented as a
dependency for whoever sequences #225 to pick up. No RBAC or manifest
behavior changed here, only the comment.

PR #899, opened 2026-08-12
chore(cluster-homelab): deploy infra-observability-core v0.29.0 with required loki secret-store key variables

ppat:
Deploys `infra-observability-core` **v0.29.0** to `homelab` and supplies the cluster-side
configuration it now requires.

Refs `ppat/homelab-ops-kubernetes-apps#3611`, `#3613`.

The Prometheus PV adoption that was originally bundled here has been split out to **#900** — it is
unrelated to this rollout and carries a different risk profile.

## BREAKING — three variables are now required

v0.29.0 includes apps#3618, which makes Loki's S3 secret-store key names **required post-build
variables with no defaults**. A consumer that does not supply them gets an ExternalSecret with empty
`remoteRef.key` values, and Loki fails. This PR supplies them with the literals the module used
previously, so **behaviour is unchanged**:

| variable | value |
| --- | --- |
| `loki_s3_endpoint_key` | `cluster_homelab_minio_loki_endpoint` |
| `loki_s3_accesskeyid_key` | `cluster_homelab_minio_loki_accesskeyid` |
| `loki_s3_secretkey_key` | `cluster_homelab_minio_loki_secretkey` |

The defaults were removed upstream deliberately: a default of `cluster_homelab_minio_loki_endpoint`
is a cluster-specific value baked into a module, which is exactly the coupling the change set out to
remove.

## `loki_s3_region` is deliberately NOT set

v0.29.0 also adds `region: ${loki_s3_region:=}` (apps#3628). Loki's S3 client signs every request
with the literal placeholder region `"dummy"` when unset (grafana/loki#11453). MinIO **ignores** the
signed region, so this has been invisible; Garage enforces an exact match and rejects with
`AuthorizationHeaderMalformed`, crashlooping the compactor.

Loki still points at MinIO, so the empty default is today's behaviour and correct. **This gets set
only when Loki cuts over to Garage** — please don't add it earlier.

## Prometheus

- `prometheus_retention_period` `336h` → **`720h`**, to support a 30-day MinIO-vs-Garage comparison.
- `prometheus_memory_request: 2Gi`, `prometheus_memory_limit: 4Gi`.

The memory raise is not cosmetic: measured peak was **1.9975 GiB against a 2 GiB limit — 99.9%** —
and a 30-day range query over ~349k series would OOMKill it. Tracked as apps#3613.

**Storage is deliberately unchanged.** `prometheus_tsdb_storage_blocks_bytes` measured 15.7–15.9 GiB
at 336h, projecting to ~34 GiB at 720h — inside both the configured 50 GiB `prometheus_retention_size`
and the 60Gi PV. Recorded because it is work deliberately *not* done.

## Note on the kube-prometheus-stack major

v0.28.0 carried a breaking chart major (87.21.0 → 88.3.0, apps#3609). That was rolled out separately
via #898 and verified live before this PR: `HelmRelease/kube-prometheus-stack-release` revision 43,
chart 88.3.0, `UpgradeSucceeded`, `Ready=True`, all `monitoring` pods Running with 0 restarts. **This
PR does not carry it** — it takes the module the rest of the way to v0.29.0.

### DIFF
--- a/clusters/homelab/kustomizations/infra-observability-core.yaml
+++ b/clusters/homelab/kustomizations/infra-observability-core.yaml
@@ -37,10 +37,22 @@ spec:
       #   prometheus_retention_size = 70-80% x prometheus_storage_size
       # must match: should match '(^0|([0-9]*[.])?[0-9]+((K|M|G|T|E|P)i?)?B)$'
       prometheus_retention_size: 50GiB
-      prometheus_retention_period: 336h
+      # 720h (30d) so the Garage-vs-MinIO trial (apps#3611) has a comparable 30-day
+      # metrics window. Storage headroom checked, not assumed: prometheus_tsdb_storage_blocks_bytes
+      # measured ~15.8GiB at 336h (2026-08-12), which projects to ~34GiB at 720h -- comfortably
+      # inside both prometheus_retention_size (50GiB) and the 60Gi PV below, so storage is
+      # unchanged here. See apps#3613 for why memory (not storage) is this cluster's actual
+      # constraint at 30d retention.
+      prometheus_retention_period: 720h
       prometheus_storage_class: sc-longhorn-replicated
       prometheus_storage_size: 60Gi
       prometheus_volume_name: pv-prometheus
+      # Pre-existing defect (apps#3613): measured 2026-08-12 at 99.9% of the module's
+      # previously-hardcoded 2Gi limit (max_over_time(container_memory_working_set_bytes{
+      # container="prometheus"}[14d]) = 1.9975GiB), and a 30-day range query over ~349k head
+      # series risks an OOMKill at that margin.
+      prometheus_memory_request: 2Gi
+      prometheus_memory_limit: 4Gi
       loki_storage_class: sc-longhorn-replicated
       loki_retention_size: 30d
       # each of the 3 monolithic replicas combines the write/read/backend roles that SimpleScalable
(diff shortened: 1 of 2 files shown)


ppat:
## `diff (homelab, helmrelease)` is failing on an inherited defect, not on this PR

Recorded so it isn't re-diagnosed. The job fails twice in a row, so it is not flaky — but the cause is **not in this PR's diff**.

Real error (the first run's `TimeoutError` masked it; the re-run surfaced it):

```
flux build ks infra-clusterops-core --dry-run --path /github/workspace/before/infrastructure/subsystems/clusterops-core
  → kustomize build failed: accumulating resources from
    'https://github.com/rancher/system-upgrade-controller/releases/download/v0.20.1/system-upgrade-controller.yaml':
    URL is a git repository
```

Three things establish this is inherited:

1. It fails while rendering the **`before/`** tree — i.e. `main` as it stands, not this PR's result.
2. It fails in **`infra-clusterops-core`**, a subsystem this PR does not touch. This PR changes only `infra-observability-core` plus a Prometheus PV.
3. `main` pins `infra-clusterops-core-v0.3.2`, which carries the unresolvable remote URL. **#891** and **#892** bump it to `v0.3.4`, and both are non-draft with all 9 checks green — and they pass this very job, because they change the pin so the `after` tree renders cleanly.

**Unblock**: land #891 and #892 first. This PR's `diff` job should then go green with no change here. If it does not, the diagnosis above is wrong and worth re-opening.

Everything else on this PR is green: `pre-commit`, `kubernetes-manifests`, `yamllint`, `commit-messages`, `alloy`, `affected-clusters`, and `diff (homelab, kustomization)`.

PR #910, opened 2026-08-13
feat(cluster-homelab): add MinIO to Garage bucket migration runbook and mechanism

ppat:
## Summary

Adds the copy mechanism, tiered verification, and rollback drill for the actual
MinIO→Garage bucket migration (apps#3611) — a manual, occasional, outside-Flux
procedure, documented as a runbook in `OPERATIONS.md` with supporting scripts under
`clusters/homelab/scripts/minio-garage-migration/` (not Flux-managed, no
`kustomization.yaml`, loaded as a ConfigMap by hand per the runbook's usage block).

This was originally opened against `ppat/garage-falsification-harness` — wrong repo.
That repo is for the falsification harness only; this is a different category of thing
(the actual copy mechanism, not a falsification test) and belongs here, which already
documents exactly this class of procedure ("a manual, occasional procedure outside
Flux") with established runbook precedent (`Runbook: prepare a node for KubeVirt`,
`Runbook: verify the sandbox NetworkPolicy falsifiability probes`) that this PR matches
in structure rather than inventing a new shape. The original PR is closed with a
pointer here; the work itself is unchanged, only the location and framing.

Production (`homelab`) is read-only and Garage isn't reachable on any real cluster yet
(its Ingress is currently patched-deleted on `homelab` pending apps#3639/clusters#908),
so this is a mechanism proven on `sandbox-talos` against synthetic data at realistic
shape (95.4% of objects under 1KB, prefix-clustered keyspace, a versioned source bucket
carrying delete markers and noncurrent versions), not a completed migration.

## The mechanism

`rclone copy` in current-versions-only mode (no `--s3-versions`), which is what makes
the migration double as the delete-marker/noncurrent-version garbage collection this
project needs (~44GB → ~13.5GB on a versioned MinIO bucket carrying 3.25M delete
markers). Resumability comes from rclone's own idempotent compare-before-transfer
behavior plus atomic single-object PUTs — deliberately not a custom checkpoint file,
given this project has already lost experiment data three times to interrupted
long-running processes with exactly that failure shape.

**The moving delta** (Loki writes to `homelab-loki-chunks` continuously): copy →
converge (bounded by write volume, not bucket size) → cutover (flip which Bitwarden key
Loki's storage config reads from — apps#3618's parameterized secret-store key names,
built for exactly this) → one final catch-up pass. Two alternatives rejected in
`OPERATIONS.md`: flip-writes-first-backfill-behind and exploit-retention-migrate-nothing
both reintroduce the dual-object-store read path apps#3611 explicitly documented as
*unverified at this Loki chart version* and chose the full-copy path specifically to
avoid.

## The two strongest pieces of evidence (called out explicitly, not buried)

1. **The rollback drill actually caught a real instance of the moving-delta problem.**
   Drilled end to end on `sandbox-talos` against a real, throwaway `grafana/loki:3.7.6`
   instance reading data this mechanism migrated — not simulated: cut over (38s), roll
   back (2s), cut over again (25s), correct query results at every step. Mid-drill, the
   first post-cutover query came back **empty** — not because Garage was missing data,
   but because Loki had shipped a tsdb index file to MinIO *after* the bulk-copy
   snapshot, and it wasn't on Garage until a catch-up pass ran. That's the drill doing
   its job rather than confirming a belief already written down, and it's the concrete
   reason the converge-then-catch-up step exists in the design rather than a single pass.
2. **Every verification tier was proven able to fail before being trusted to pass.**
   This project voided an entire experiment once (H4, apps#3611) on a check that could
   not have failed by construction — demonstrated-red is the standard here, not "it
   copied without errors":
   - Version exclusion: an independent oracle (`ListObjectVersions` filtered to
     `IsLatest && !DeleteMarker`, a different API call than the tool's own listing) said
     1,425 keys should exist current. A deliberately wrong naive copy produced **1,500**
     — correctly flagged as wrong. The real `rclone copy` produced exactly **1,425**.
   - Corruption detection: tier2 passed pre-corruption, **failed** on an injected
     same-size random overwrite (confirmed genuinely hash-based via rclone's own `md5
     differ` log line, not size-only), passed again after repair.
   - Resumability: a live copy was SIGKILLed at **582/1,425** objects transferred,
     resumed with the identical command, and the final destination state was
     byte-identical (full hash comparison) to an uninterrupted baseline, zero truncated
     objects.

## What remains assumption, stated plainly

- **Behavior at the real 3.44M-object count.** Prototyped at ~28,500–30,000 objects per
  run. The project's own H1 falsification harness already exhaustively measured both
  engines' behavior at 3.44M objects (RSS, LIST p99, disk usage) — this didn't need to
  duplicate that; what needed proving here was the *copy mechanism's* correctness, which
  object count doesn't change the shape of. The 3.44M-object convergence timeline itself
  (pass count, final-gap duration at real scale) is consequently not measured either.
- **Production Loki's actual rollback cost.** The drill used a minimal single-binary
  Loki with no caches and one replica; production Loki (`loki-chunks-cache`/
  `loki-results-cache` sidecars, more replicas) will have a different restart cost than
  the measured 2–38s — the *mechanism* doesn't change, the number will.
- **Real bucket contents.** Synthetic fixture data matches the stated shape but is not
  `homelab-loki-chunks`'s actual data — production is read-only.

## Munger's razor applied

`homelab-authentik-media` (5 objects) and `homelab-terraform-state` (7 objects) get the
same script, run once, by hand — not a shrunk-down convergence/sampling pipeline.
Proven on sandbox: both copied and fully verified in under a second each.

## Out of scope, flagged not solved

Migrating `homelab-terraform-state`'s Terraform *backend config* (six workspaces'
`backend "s3"` blocks) is separate, deliberate follow-on work — copying the object
doesn't switch which bucket Terraform reads from next, and that bucket holds the state
of the MinIO workspaces themselves (a chicken-and-egg this doesn't make worse but also
doesn't solve).

## Facts verified against live repo state, not assumed

- apps#3618 (parameterized Loki S3 secret-store key names) is merged; confirmed live in
  `clusters/homelab/kustomizations/infra-observability-core.yaml` at time of writing.
- `garage_s3_region` is a required `postBuild` variable on `infra-storage-core`,
  currently `us-east-1`-convention per the module's own comments — `loki_s3_region` must
  match exactly at cutover, checked at pre-flight time per the runbook (Garage rejects a
  mismatch with `AuthorizationHeaderMalformed` on every request, no client-side retry,
  already reproduced live against `sandbox-talos`'s `logging/loki-0`, 278+ restarts).
- Garage's own Ingress (`garage-s3`) is **currently patched-deleted** on `homelab`
  (`infra-storage-core.yaml`'s patch list, neutralizing the reverted operator-based
  deploy) — there is no live, reachable Garage endpoint on this cluster today. This
  runbook is written for use once apps#3639/clusters#908's plain-resources rebuild lands
  and that Ingress is live again, not for immediate execution.

## Test plan

- [x] `pre-commit run` (yamllint, shellcheck, markdownlint, commitlint,
      check-shebang-scripts-are-executable) clean on all new/changed files
- [x] All four falsifiability tests + the full four-bucket sequence + the rollback drill
      run live on `sandbox-talos` against real MinIO (`RELEASE.2025-04-22T22-12-26Z`,
      the exact production build) and real Garage (`dxflrs/garage:v2.3.0`, same) — exact
      numbers in `OPERATIONS.md`
- [ ] Owner review of the rejected delta-handling alternatives and the `services/` vs.
      plain-directory placement decision for non-Flux-managed scripts


ppat:
Closing unmerged, deliberately. Recording where each half went, because both are still live — this is not abandoned work.

**The tooling was extracted and is what production will actually run.** The rehearsal's forward copy and its Phase 8 reverse copy both executed `bin/migrate-bucket.sh` and `bin/verify-bucket.sh` byte-identical, with only the rclone remote roles inverted. All twelve files (`bin/`, `k8s/`, `test/`) now live outside this repo under local version control, with the as-extracted state as the baseline commit so corrections stay diffable against the drilled version.

One correction has since been applied to that copy: `verify-bucket.sh`'s tier0 was a two-sided count-and-bytes equality test, which **reds on a correct reverse copy** — a reverse copy legitimately leaves the destination a superset. tier0 is now gated on copy direction and splits the two mismatch shapes it previously collapsed (destination short = a real defect; destination superset = tier0 has no standing). tier1 and tier2 needed nothing: both already use `rclone check --one-way`, which is direction-agnostic.

**The runbook in this PR body is superseded and should not be followed.** A 16-document plan replaced it — drilled end to end across eight phases against real production data, then corrected against four independent audits. This body stays readable as provenance for how the mechanism was reasoned about; it is no longer the procedure.

**Known defects in the code as it stands here**, all recorded against the extracted copy rather than fixed in this branch:

- `entrypoint.sh` hardcodes `SCRIPT_DIR=/bin/scripts` with no override, so it only runs inside the Job.
- The Job mounts a `migration-scripts` ConfigMap that nothing in this branch creates.
- `test/test-corruption-detection.sh` resolves its target via `$(dirname $0)` and cannot find `bin/verify-bucket.sh`.
- `bin/rclone.conf.template` is dead — `render-rclone-conf.sh` carries its own inline copy of the same stanzas.

Nothing here needs merging for the migration to proceed.

